PR #4, opened 2021-11-24
Implementation of list minimal adjustment sets (identification)

AndrewC19:
## Overview
I have implemented and tested an algorithm for listing minimal adjustment sets given a causal DAG, a set of treatment variables, and a set of outcome variables. The main algorithm used in this implementation is presented in [1] but a simpler version can be found in the earlier conference version [2]. This is the algorithm used by Dagitty which is significantly faster than DoWhy for identification. However, Dagitty is only available as an R package so to reduce dependencies it made sense to implement it for ourselves.

At a high-level, the algorithm works as follows:
1. Convert the DAG to a proper back-door graph, relative to the specified treatments (X) and outcomes (Y) [1].
2. Convert the proper back-door graph into an ancestor graph, relative to X and Y [1, 2].
3. Convert the ancestor graph into a moralised graph [1, 2].
4. Add a node XM which is connected to all variables in X [1].
5. Add a node YM which is connected to all variables in Y [1].
6. Remove all nodes in X and Y from the graph and connect their neighbours [1].
7. Apply Takata's list minimum separators algorithm [3] to the resulting graph [1, 2].
8. Return the set of minimum separators -- these are equivalent to adjustment sets in the original DAG [1].

All of the methods used in this algorithm have been implemented in the `CausalDAG` class, with the exception of two methods which only apply to the undirected graphs used in steps 2 and 3. These are currently in `causal_dag.py` but outside of the `CausalDAG` class. We can these a better home (e.g. helper functions library) in the future. All methods have also been unit tested in the `tests/specification_tests/test_causal_dag.py` achieving 96% coverage over `causal_dag.py`.

## TODO:

- [ ] Check guidelines are met (not quite finalised yet).
- [ ] Work out how to set-up continuous integration and automatic coverage report upon PR.

## References
[1] [Zander et al., Separators and adjustment sets in causal graphs: Complete criteria and an algorithmic framework, 2019](https://www.sciencedirect.com/science/article/pii/S0004370219300025?casa_token=EYziiE_8VuYAAAAA:ArtIagD0Jyh6_Hfb69RAnchlqrRaundthGdqNMpr6imoWEEZD9zAmoEOKXLopLGM6KwttCkB-wM)
[2] [Textor and Liskiewicz, Adjustment Criteria in Causal Diagrams: An Algorithmic Perspective, 2012](https://arxiv.org/abs/1202.3764)
[3] [Ken Takata, Space-optimal, backtracking algorithms to list the minimal vertex separators of a graph, 2010](https://www.sciencedirect.com/science/article/pii/S0166218X10001824)

### DIFF
--- a/causal_testing/specification/causal_dag.py
+++ b/causal_testing/specification/causal_dag.py
@@ -1,6 +1,7 @@
 import networkx as nx
 import logging
 from typing import Union
+
 Node = Union[str, int]  # Node type hint: A node is a string or an int
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


jmafoster1 (commented):
I've not downloaded and run it as I don't know how to do that without merging and pulling, bug it all looks OK. Is it formatted with Black? Is that something we want to do?

AndrewC19:
It's not formatted with black but I have tried to stick to the rough style guidelines in the contributing guidelines.

jmafoster1 (approved):

PR #16, opened 2021-11-29
Redefined Scenarios to better suit data generation

jmafoster1:
As per discussion with @AndrewC19, I've changed the `Scenario` class so that it's now a set of variables and constraints over them rather than a `{str: Constraint}` map. I've also dispensed with the `Constraint` class in favour of Z3 `ExprRef` constraints and `scipy` distributions. I've not done implementing stuff yet, but I thought it'd be good to get these changes in as soon as possible.


causal_testing/specification/causal_dag.py
```diff
@@ -6,10 +6,17 @@
 
 Node = Union[str, int]  # Node type hint: A node is a string or an int
 logger = logging.getLogger(__name__)
+from .scenario import Scenario
+from .variable import Output
 
 
-def list_all_min_sep(graph: nx.Graph, treatment_node: Node, outcome_node: Node, treatment_node_set: {Node},
-                     outcome_node_set: {Node}):
+def list_all_min_sep(
```
AndrewC19, line 13: @jmafoster1 Why are the methods formatter like this? I have always preferred the former style -- open to changing though.

causal_testing/specification/causal_specification.py
```diff
@@ -1,33 +1,43 @@
 from abc import ABC
-from causal_testing.specification.constraint import NormalDistribution
+
 from causal_testing.specification.causal_dag import CausalDAG
 from causal_testing.specification.scenario import Scenario
+
 from typing import Union
 import logging
 
+# from causal_testing.specification.scenario import Scenario
```
AndrewC19, line 9: Remove comment

causal_testing/specification/causal_specification.py
```diff
@@ -1,33 +1,43 @@
...
-    causal_dag.add_edge("Vaccine", "Cumulative Infections")
-    causal_dag.add_edge("Age", "Vaccine")
-    causal_dag.add_edge("Age", "Cumulative Infections")
-
-    causal_specification = CausalSpecification(scenario, causal_dag)
-    print(causal_specification)
+        return f"Scenario: {self.scenario}\nCausal DAG:\n{self.causal_dag}"
+
+
+# def main():
```
AndrewC19, line 24: Remove comments and if using main in script, put the main body into `if __name__ == "__main__":`

causal_testing/specification/causal_dag.py
jmafoster1: It's just how Black does it. I have no strong preference either way. It's not something I've conciously changed.

setup.py
```diff
@@ -1,18 +1,5 @@
 from setuptools import setup, find_packages
 
-requirements = [
```
AndrewC19, line 3: Why have you removed requirements?

tests/specification_tests/test_causal_dag.py
```diff
@@ -1,8 +1,15 @@
 import unittest
 import os
 import networkx as nx
-from causal_testing.specification.causal_dag import CausalDAG, close_separator, list_all_min_sep
-from tests.test_helpers import create_temp_dir_if_non_existent, remove_temp_dir_if_existent
+from causal_testing.specification.causal_dag import (
```
AndrewC19, line 4: I've never seen imports with brackets like this

causal_testing/specification/causal_specification.py
jmafoster1: Removed

AndrewC19 (changes requested):
Fix setup.py and I'm unsure of the brackets for imports

causal_testing/specification/causal_specification.py
jmafoster1: Removed

setup.py
jmafoster1: That was a bad merge. I've fixed it now

tests/specification_tests/test_causal_dag.py
jmafoster1: Again, this is Black being "uncompromising" and is not something I do conciously. It just autoformats. Perhaps we should add this to the discussion with @bobturneruk about coding style.

causal_testing/specification/causal_dag.py
AndrewC19: I would personally prefer not to use this kind of convention, it requires a lot of lines and doesn't make the code any more readable. I prefer to use the line width. Would you mind reverting these style changes?

tests/specification_tests/test_causal_dag.py
AndrewC19: Agree. Good idea. For now I think it is best to revert to how it was for continuity with the rest of the repository.

causal_testing/specification/causal_dag.py
jmafoster1: That'll mean I'll have to disable Black, which might make my code more anarchic in the long run. If you're OK with that then sure, although it might take a while to go through and revert all the methods that have been Blacked.

setup.py
```diff
@@ -1,21 +1,26 @@
...
+    "setuptools~=58.5.3",
+    "networkx~=2.6.3",
+    "pygraphviz~=1.7",
+    "pytest~=6.2.5",
+    "scikit-learn~=1.0.1",
+    "matplotlib~=3.5.0",
+    "econml~=0.12.0",
+    "statsmodels~=0.13.1",
+    # DO NOT CHANGE THE ORDER OF THESE!
+    "z3~=0.2.0",
```
bobturneruk: Suggest try without line 15.

AndrewC19 (approved):

PR #18, opened 2021-12-01
Removed redundant and old tests

AndrewC19:
I have removed failing test cases which should never have been committed originally.

These tests are replaced by @jmafoster1 in the generation branch.


bobturneruk (approved):
Looks good. As I don't know exactly what the software is supposed to do, I can't comment on whether a specific test is appropriate or not. I think @AndrewC19 's opinion is fine to go with at this point!

PR #19, opened 2021-12-03
Add estimators

AndrewC19:
I have added a basic implementation for the estimator class, including a linear regression estimator using statsmodels and a causal forest estimator using EconML. These are not intended to be plug-and-play estimators, but examples that show how a user can implement the estimator class. For most non-trivial use cases, I think we can expect that the user will want to at least tweak the estimators. 

Both estimators have been tested against examples from (Hernán and Robins, Causal Inference: What If?, 2020). However, there are a few limitations at this point:
- I have not used the linear regression estimator for conditional average treatment effects yet (not absolutely necessary and a user could do this if needed). I think this would be easier using a linear EconML model since these are designed for heterogeneous treatment effects.
- I have not tested the causal forest estimator thoroughly yet. This is because the estimate changes quite a lot each time you run the estimator (it uses a lot of ML), so we can't compare it to the outcomes in the book (these use a linear regression estimate which is stable). We should look into the most appropriate way to test ML-heavy estimators like this.


AndrewC19:
@bobturneruk one of the dependencies of EconM, llvmlite (https://github.com/numba/llvmlite), isn't compatible with Python 3.10. Any ideas what the best way forwards with this is?

bobturneruk:
Perhaps we should not support `3.10` yet? If not, edit the workflow https://github.com/CITCOM-project/CausalTestingFramework/blob/main/.github/workflows/ci-tests.yaml 

See here for active versions:
https://www.python.org/downloads/

jmafoster1 (approved):
For the linear regression, it might be more generally applicable to have an `add_power_term_to_df` function in which the user specifies a term and a power rather than just `add_squared_term_to_df`. Then we can easily do arbitrary powers. Otherwise looks good.

PR #27, opened 2021-12-13
Refactor causal test engine

jmafoster1:
Moved a bunch of stuff around! Most notably, I've put `abstract_causal_test_case.py`  in its own subpackage called `generation` to minimise ambiguity. I've also tidied up a bunch of stuff.


causal_testing/data_collection/data_collector.py
```diff
@@ -61,62 +120,22 @@ def run_system_with_input_configuration(
...
+        data which is invalid for the scenario-under-test. Data is invalid if it does not meet the constraints
         outlined in the scenario-under-test (Scenario).
+
+        :param scenario: Scenario for which the observational data is collected.
         :param csv_path: Path to the csv containing execution data.
         :return scenario_execution_data_df: A pandas dataframe containing execution data that is valid for the
         scenario-under-test.
         """
-        execution_data_df = pd.read_csv(csv_path)
+        execution_data_df = pd.read_csv(self.csv_path, **kwargs)
```
AndrewC19, line 139: I am not sure this is a good way to use **kwargs. If you are using the **kwargs specifically to pass args to read_csv, it might be better to have them as explicit keyword arguments.

jmafoster1: Fair enough. I was mainly just keen to avoid duplicating all the kwargs to `pd.read_csv`. The only one I actually needed was `index_col`, so I just added support for that and removed `**kwargs`.

jmafoster1: Update: Turns out codacity didn't like that, so I had to revert back :|

AndrewC19 (approved):

PR #29, opened 2021-12-15
Covid 19 example

AndrewC19:
This PR adds a simple causal testing example based on a Covasim simulation.

Summary of key changes:

1. I have added the `examples` directory that we can fill with examples of causal testing.
2. I have added the `examples/covasim_` directory for this example, containing a DAG (dot file and png), some observational data, and a script that creates and executes the causal test case.
3. A few minor changes to components of the causal testing framework that I noticed during development of the example (e.g. adding getters to `CausalTestCase` for easier access to treatment and outcome variables and control and treatment values).

TODOs (issues to add):

- [ ] This example shows that the current process for creating and executing test cases is quite clunky. There are a lot of steps that are repetitive and could probably be automated. We should now look to pinpoint the steps that concern the user and those that can be handled behind the scenes.
- [ ] Currently, `CausalTestEngine` works with a single test case. We need to make it work with a collection of test cases so that  a user can specify a number of tests for each scenario and execute them all at once. One problem we will have to overcome is how to handle estimator specification when there are multiple test cases for one scenario. The current workflow has the user (1) instantiate a `CausalTestEngine` instance, (2) run `load_data`to either generate the execution data or to filter user supplied observational data and return the adjustment set, (3) specify an estimator (using the adjustment set), (4) execute the test case with the estimator `execute_test(estimator, 'ate')`. We need to think of a clean way to conduct this workflow when there are multiple test cases.


AndrewC19:
I have also changed the design of causal test engine so that it takes a data collector as an argument. This means that we can handle data collection uniformally (i.e. just call `data_collector.collect_data(**kwargs)` instead of having a case for each type of data collector. Also, it makes sense for the user to specify the data collector upfront since, in the experimental case, they must implement the methods in `ExperimentalDataCollector` to execute their system.

jmafoster1 (approved):

PR #45, opened 2022-02-25
Sphinx docs

cwild-UoS:
Further doc information would be great but they are in a good place to be merged.

Further pages could include a simple user guide, examples, detailed theory etc.


cwild-UoS:
> Good to see you've won the battle with this @cwild-UoS. A few points (always the possibility I'm doing something daft!):
> 
> * Include local build instructions in readme (e.g. `pip install sphinx`, `cd docs`, `make html`).
> * My local build (Win10, conda) looked different to what you last demo'd (is that online somewhere?). Do I need a flag on the `make` command?
> * `.gitignore` docs outputs (e.g. `*.htm*`).
> * Go with a single `conf.py` if possible (there's on in the root and `docs\source\conf.py`).

A readme is a good idea, I've added one now which points to the current readthedocs location and how to build locally.

If you're local build doesn't look like the readthedocs, that would be very interesting and we can discuss it in our catchup tomorrow? 

In regards to the `.gitignore`, I have ignored the build folder: 
https://github.com/CITCOM-project/CausalTestingFramework/blob/sphinx-docs/.gitignore#L57
Should I also include the `*.htm*` files as well? 

`/docs/source/conf.py` is necessary for the setup of the docs building. However `.readthedocs.yaml` in the root directory is there to tell readthedocs how to setup a environment in which to then build the docs in. We can get rid of it by doing all the settings in the readthedocs settings gui for the project. However it will no longer be as visible and wont be including in the VC. Let me know what you think about this trade-off

bobturneruk (approved):
This'll be good once the spare `conf.py` has gone, I think!

jmafoster1 (approved):

PR #50, opened 2022-03-17
Converted type annotations of any type to Any type

jmafoster1:
As per discussion [here](https://github.com/CITCOM-project/CausalTestingFramework/pull/39).


causal_testing/testing/estimators.py
```diff
@@ -71,7 +72,7 @@ class LinearRegressionEstimator(Estimator):
     combination of parameters and functions of the variables (note these functions need not be linear).
     """
     def __init__(self, treatment: tuple, treatment_values: float, control_values: float, adjustment_set: set,
-                 outcome: tuple, df: pd.DataFrame = None, effect_modifiers: {Variable: any} = None, product_terms: [(Variable, Variable)] = None):
```
cwild-UoS, line 74: Rather than make a List of typing's, I think it is preferred to define with typing that it is a list:

 `product_terms: List[Tuple[Variable, Variable]]` or `product_terms: Sequence[Tuple[Variable, Variable]]`

cwild-UoS (approved):

causal_testing/testing/estimators.py
jmafoster1: This is my inner functional programmer coming out again as this is Haskell syntax! The "preferred" syntax looks more like Scala, so I'll switch to that. Is there any easy way to automatically fix it, or shall I just merge in and gradually change the ones I see as I come across them?

jmafoster1: If you're happy to do that, I'm happy for you to do that. I can't say the idea of going through code and checking datatypes sparks joy... As you say, it might generate more typing questions that we can get resolved.

PR #57, opened 2022-04-13
Merge case study progress as examples

AndrewC19:
This PR adds the latest version of the case studies as example applications of the Causal Testing Framework.

There are two case studies:
- One for the Luo-Rudy 1991 Caridovascular AP Model
- One for a Covasim modelling scenario

There should be no changes to the Causal Testing Framework itself, just the examples and necessary csv data to run them.


bobturneruk:
Good plan. I'd be keen to have some input on this as it takes shape. I like the consistent folder structure and separation of data. Good to think about tests i.e. running the examples in the CI to make sure changes to the framework don't break them. Not sure about the `bessemer` folder.

We did this sort of thing for examples before on pykale (https://github.com/pykale/pykale/tree/main/examples). You might like to look at how we did notebooks using colab and mybinder (https://github.com/pykale/pykale#step-1-tutorials-and-examples, pykale: [Open in Colab](https://colab.research.google.com/github/pykale/pykale/blob/main/examples/digits_dann_lightn/tutorial.ipynb)). I did the first one and the project team followed on with more.

AndrewC19:
Hi @bobturneruk, your input on the case studies would be greatly appreciated. There are certainly better ways to present them that I'd be interested in looking at e.g. colab and mybinder.

I added the bessemer directory by accident so I'll remove that. I only want to include the csv files that are absolutely necessary to reproduce the results.

jmafoster1 (approved):

PR #67, opened 2022-05-18
Json poisson example

cwild-UoS:
I think a discussion around the usability of https://github.com/CITCOM-project/CausalTestingFramework/blob/json_poisson_example/json_frontend/examples/poisson/causal_test_setup.py would be useful.

Currently the frontend assumes you correctly structure the custom user python code as per the example which isn't ideal. The obvious alternative to me is having a class that can be extended, maybe even an abstract one with properties to enforce it is done correctly. However if we want the framework to be as open as possible to people without python background classes and class extensions might just be another barrier to using it?

Not sure if this is best discussed in a comment or two or with a meeting :)


jmafoster1:
Personally, I quite like a class extension, but it is potentially a "barrier to entry". Having said that, if someone can write a complex computational model which requires our testing framework in order to test it effectively, they should be able to manage extending a class...

cwild-UoS:
That is probably very true, if they can jump every barrier up to the point of using the framework, one class extension hopefully won't stop them!

I'll create the class extension and then if needed, we can always discuss again when it's in front of us

jmafoster1:
Go ahead. I have no strong opinions...

cwild-UoS:
> How would it be if https://github.com/CITCOM-project/CausalTestingFramework/blob/json_poisson_example/json_frontend/process_causal_tests_json.py were moved into `causal_testing/`, https://github.com/CITCOM-project/CausalTestingFramework/tree/json_poisson_example/json_frontend/examples/poisson moved into `examples/` and the readmes merged? Would this better present the "JSON Frontend" as part of causal testing framework? cc @jmafoster1 @cwild-UoS @AndrewC19

Sounds good to me @bobturneruk, it could be overlooked in its current structure

cwild-UoS:
Very unhelpful stack trace I agree..

I just tried again with a fresh conda environment with the latest CTF from main branch and it worked okay. When did you last update the CTF in your conda env?

jmafoster1:
Thursday

causal_testing/json_front/README.md
```diff
@@ -0,0 +1,32 @@
+# JSON Causal Testing Framework Frontend
```
bobturneruk, line 1: This file is going to be a bit tricky to find here in the package. It might be better merged with the main readme for now.

causal_testing/json_front/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+`examples/poisson/causal_test_setup.py` contains python code written by the user to implement scenario specific features
+such as:
+1. Custom Estimators
+2. Causal Variable specification
+3. Causal test case outcomes
+4. Meta constraint functions
+5. Mapping JSON distributions, effects, and estimators to python objects
+
+`examples/poisson/causal_tests.json` is the JSON file that allows for the easy specification of multiple causal tests.
```
bobturneruk, line 17: We'll need quite a bit more instruction for anyone not already familiar with the package to use this. I suggest making an issue to keep track of this. It could be that a jupyter notebook is is a good next step.

examples/poisson/causal_test_setup.py
```diff
@@ -0,0 +1,197 @@
...
+def get_args() -> argparse.Namespace:
+    """ Command-line arguments
+
+    :return: parsed command line arguments
+    """
+    parser = argparse.ArgumentParser(
+        description="A script for parsing json config files for the Causal Testing Framework")
+    parser.add_argument("-f", help="if included, the script will stop if a test fails",
+                        action="store_true")
+    parser.add_argument("--directory_path", help="path to the json file containing the causal tests",
```
bobturneruk, line 107: I'm not sure the default will work here, as it's supposed to be setting a folder / directory, not a specific file. And the filenames are hardcoded elsewhere (in `causal_testing/json_front/json_class.py`). Maybe get rid of the CLI arg and look for files in the same folder as the script for now? Or get rid of the default value? We may want to look at this again after merging, depending on expected user needs around file locations and naming. I would have thought that, at least, not everyone would want to call their data `data.csv` which I think is the current requirement.

bobturneruk (approved):
I think this looks good. I'd like to get a deeper understanding and maybe a simpler example / more documentation / a jupyter notebook implementation will help with this once merged. I wonder if some things might be named differently. Maybe "json interface" is better than "json frontend"? And perhaps `causal_test_setup.py` could be `run_causal_tests.py`?

Generally liking how all the example specific stuff is in `examples/poisson` now.

setup.py
```diff
@@ -14,12 +14,18 @@
...
 ]
 
 setup(
     name="causal_testing_framework",
     version="0.0.1",
     install_requires=requirements,
     packages=find_packages(),
+    entry_points={
+        'console_scripts': [
+            'run_json = causal_testing.json_frontend:main'
```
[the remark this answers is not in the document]
cwild-UoS: Thanks for spotting this, I was using this method before we swapped the main function from being in `json_class.py` to `causal_test_setup.py`. Now that the main function is in a script that the user is free to put and call from where they want, I'm not sure if there is a way to have an entry point to the main function. 

Ill remove these lines for now

examples/poisson/causal_test_setup.py
```diff
@@ -0,0 +1,197 @@
...
+            self.adjustment_set.add(k)
+            x[k] = v
+        if hasattr(self, "product_terms"):
+            for a, b in self.product_terms:
+                x[f"{a}*{b}"] = x[a] * x[b]
+
+        x.drop(['width', 'height'], axis=1, inplace=True)
+        self.adjustment_set = {"width*intensity", "height*intensity"}
+
+        print(x)
```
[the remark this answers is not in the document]
cwild-UoS: Definitely a priority :)

examples/poisson/causal_test_setup.py
cwild-UoS: You're right. I think now that the door has been opened in terms of the user writing more python code, the simplest approach is expect the user to specify the paths in `causal_test_setup.py`. This way the file names and locations are completely flexible and up to the user

causal_testing/json_front/README.md
cwild-UoS: Would it be a good idea to move this readme to the readthedocs and point to it from the main readme?

bobturneruk: That would also work. Examples / tutorials should be prominent.

Issue #37, opened 2022-01-25
Remove `requirements.txt`


bobturneruk:
`requirements.txt` is now redundant given we have a separate list of packages in `setup.py`.

PR #70, opened 2022-05-19
Removed requirements.txt

AndrewC19:
Removed redundant requirements.txt as per #37.


bobturneruk (approved):
Looks good. Requirements have been in `setup.py` for a while so happy to see thins go. Thanks!

PR #73, opened 2022-07-12
Updated installation instructions in readme

jmafoster1:
Tried to install to my new office machine and had a couple of (easily resolvable) issues, so I thought I'd update the installation instructions accordingly.


cwild-UoS (approved):
Not very familiar with GNU and its compilers. Is this a linux/debian based requirement only? Or will people on windows and mac need to find equivalent instructions?

jmafoster1:
Yeah this is a linux thing. I'd assumed that was clear from the previous sentence, but if not, I can add extra clarification?

cwild-UoS:
No you're right, ignore me! I think it looks good and I was just overthinking things :)

PR #78, opened 2022-07-26
add logging to stdout and log file

cwild-UoS:


jmafoster1 (commented):
Would you mind fixing the codacity errors before merging please? There's only 4 or 5 and they're all pretty straightforward. I'm not sure about the codacity stuff. It might be good to have some tests for this, especially if we can capture the logger output and assert that particular tests should be passing/failing

cwild-UoS:
> Would you mind fixing the codacity errors before merging please? There's only 4 or 5 and they're all pretty straightforward. I'm not sure about the codacity stuff. 

Yep no worries, it was on the to do list :)!

>It might be good to have some tests for this, especially if we can capture the logger output and assert that particular tests should be passing/failing

Because I didn't add tests for any of the PRs relating to the json frontend, my next objective is to go through and get the coverage up across the entire module hopefully. Some particular-test tests does sound like a good idea though.

bobturneruk:
Good to hear. Often it's easiest to write tests around the same time as the code..

jmafoster1 (approved):

PR #91, opened 2022-08-16
Fix minimal adjustment set

### LINKED ISSUE #90: Minimal Adjustment Sets
Used Dagitty to find the minimal adjustment set of the DAG we are using. Daggity said the minimal adjustment set for the node 'weight' was the node 'content', however the CTF said it was the node 'S3'.
`digraph DAG {
    rankdir=LR;
    content -> weight;
    weight -> S3;
    country -> S3;
    country -> distance;
    content -> S3;
    plane_transport -> S1;
    plane_transport -> S2;
    S1 -> alarm;
    S2 -> alarm;
    S3 -> alarm;
}
`
DAG from Daggity
image.

When we print out the Minimal Adjustment Set, it prints out both S3 and Content and it picks S3 as it is first. Daggity said the only one it should be is 'Content'.

jmafoster1:
Closes #90

### DIFF
--- a/causal_testing/specification/causal_dag.py
+++ b/causal_testing/specification/causal_dag.py
@@ -311,9 +311,11 @@ class CausalDAG(nx.DiGraph):
                 outcome_node_set,
             )
         )
-        valid_minimum_adjustment_sets = [adj for adj in minimum_adjustment_sets
-                                         if self.constructive_backdoor_criterion(proper_backdoor_graph, treatments,
-                                                                                 outcomes, adj)]
+        valid_minimum_adjustment_sets = [
+            adj
+            for adj in minimum_adjustment_sets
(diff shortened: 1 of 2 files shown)


bobturneruk:
Can you run `black` to standardise formatting, please?

jmafoster1:
Done. Sorry, I thought it did it automatically when you pushed. Is that a thing we can have it do?

bobturneruk:
Yes - megalinter would need to make a commit here (you'd need to pull), or we could set it up as a pre-commit (which would run locally).

jmafoster1:
Perhaps one to talk about in the next CITCoM update, unless one is objectively better than the other

AndrewC19 (approved):

PR #99, opened 2022-10-12
Linting Updates

cwild-UoS:
Originally was going to solve more of the linting issues but I think it would be more appropriate on the other side of the discussed refactor

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -151,7 +151,8 @@ disable=raw-checker-failed,
         suppressed-message,
         useless-suppression,
         deprecated-pragma,
-        use-symbolic-message-instead
+        use-symbolic-message-instead,
+        invalid-name
 
 # Enable the message, report, category or checker with the given id(s). You can
 # either give multiple identifier separated by comma (,) or put this option
--- a/.pylintrc
+++ b/.pylintrc
@@ -152,7 +152,8 @@ disable=raw-checker-failed,
         useless-suppression,
         deprecated-pragma,
         use-symbolic-message-instead,
-        invalid-name
+        invalid-name,
+        unnecessary-pass
 
 # Enable the message, report, category or checker with the given id(s). You can
 # either give multiple identifier separated by comma (,) or put this option
--- a/.pylintrc
+++ b/.pylintrc
@@ -152,7 +152,6 @@ disable=raw-checker-failed,
         useless-suppression,
         deprecated-pragma,
         use-symbolic-message-instead,
-        invalid-name
 
 # Enable the message, report, category or checker with the given id(s). You can
 # either give multiple identifier separated by comma (,) or put this option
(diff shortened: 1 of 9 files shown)


.pylintrc
```diff
@@ -151,7 +151,9 @@ disable=raw-checker-failed,
         suppressed-message,
         useless-suppression,
         deprecated-pragma,
-        use-symbolic-message-instead
+        use-symbolic-message-instead,
+        invalid-name,
```
bobturneruk: It is possible I think to set this up to check for either snake_case or a single character string, which I think would cover all of our variables, via a regex. Although I'm not sure why the single characters are being complained about by pylint as I think single characters *are* valid snake case.

.pylintrc
```diff
@@ -151,7 +151,9 @@ disable=raw-checker-failed,
...
+        unnecessary-pass
```
bobturneruk: We don't need this - I think some pass statements are necessary, some aren't. pylint can tell. I've pushed a commit ([001fedc](https://github.com/CITCOM-project/CausalTestingFramework/pull/99/commits/001fedc3d448489b36a1f29c20506e7b9e3b463a)) so hopefully I'm not wrong.

causal_testing/data_collection/data_collector.py
```diff
@@ -73,7 +73,10 @@ def filter_valid_data(self, data: pd.DataFrame, check_pos: bool = True) -> pd.Da
         size_diff = len(data) - len(satisfying_data)
         if size_diff > 0:
             logger.warning(
-                f"Discarded {size_diff}/{len(data)} values due to constraint violations.\n" f"For example{unsat_core}"
+                "Discarded %s/%s values due to constraint violations.\n" "For example%s",
```
bobturneruk, line 75: Looking at these [w1203](https://pycodequ.al/docs/pylint-messages/w1203-logging-fstring-interpolation.html) things, I prefer them as F-strings. I think the readability of F-strings outweighs the performance / other potential benefits of %. More discussion here https://github.com/PyCQA/pylint/issues/2395

bobturneruk: This would mean disabling some checks for a clean linter run.

bobturneruk (approved):
Thanks for looking at this - I have a few suggestions.

causal_testing/data_collection/data_collector.py
cwild-UoS: I agree as well, F-Strings are much clearer and I'm not sure there would be a noticeable performance impact. So I'll happily revert them to F-Strings for now!

Interesting read as well cheers Bob :)

.pylintrc
cwild-UoS: Ahh okay, I see that it is unnecessary when there are docstrings present as it's not 'empty code'. Personally I like them as they imply that the method is incomplete / needs filling in. I'm happy either way and to go with what the linter wants.

.pylintrc
cwild-UoS: I've added a regex to catch snake_case of 1 or more character variable names now

PR #103, opened 2022-11-21
Base causal test case

cwild-UoS:
Draft PR for visibility on changes.

Even though its not ready for review please do chip in and comment if there is anything you would like to add / discuss!


cwild-UoS:
List of changes within this PR:

1. Added `BaseTestCase` data class, which provides all the information required to perform identification and is used to instantiate a `CausalTestCase` object. 
2. Move `Identification` from the `CausalTestEngine` to the `CausalDag` class. This is because information is no longer needed from the `TestEngine` and requires an engine to be instantiated when you want to perform identification.
3. Identification now only requires a `BaseTestCase` to be passed to it to run within a `CausalDag` class. 
4. Add the `CausalTestSuite` class, which provides a structure to specify multiple `CausalTestCases` against a single `BaseTestCase`, this allows for the easy iteration over a bunch of tests against various `treatment` and `outcome` variables as well as varying `Estimators`. 
5. Add the `execute_test_suite` function to the `causal_test_engine` to execute `test_suites`
6. Update Json frontend, `AbstractTestCases` and other parts of the CTF to run with the other changes

causal_testing/json_front/json_class.py
```diff
@@ -89,10 +89,11 @@ def setup(self):
         self._populate_metas()
 
     def _create_abstract_test_case(self, test, mutates, effects):
+
         abstract_test = AbstractCausalTestCase(
             scenario=self.modelling_scenario,
             intervention_constraints=[mutates[v](k) for k, v in test["mutations"].items()],
-            treatment_variables={self.modelling_scenario.variables[v] for v in test["mutations"]},
+            treatment_variables=self.modelling_scenario.variables[next(iter(test["mutations"]))],
```
jmafoster1, line 96: What's this line doing? Could we not just do `self.modelling_scenario.variables[test["mutations"][0]]`?

causal_testing/specification/causal_dag.py
```diff
@@ -465,5 +465,28 @@ def depends_on_outputs(self, node: Node, scenario: Scenario) -> bool:
             return True
         return any([self.depends_on_outputs(n, scenario) for n in self.graph.predecessors(node)])
 
+    def identification(self, base_test_case):
```
jmafoster1, line 468: Is there a reason why we need to pass in `base_test_case` rather than just the effect we want to identify (i.e. "total" or "direct")? Is it better practice for some reason or just a personal choice?

causal_testing/testing/causal_test_case.py
```diff
@@ -3,12 +3,13 @@
...
 from causal_testing.specification.variable import Variable
 from causal_testing.testing.causal_test_outcome import CausalTestOutcome
+from causal_testing.testing.base_test_case import BaseTestCase
 
 logger = logging.getLogger(__name__)
 
 
 class CausalTestCase:
-    """
+    """)
```
jmafoster1, line 12: Is this deliberate or a typo?

causal_testing/testing/causal_test_case.py
```diff
@@ -56,7 +57,7 @@ def get_treatment_variables(self):
 
     def get_outcome_variables(self):
         """Return a list of the outcome variables (as strings) for this causal test case."""
-        return [v.name for v in self.outcome_variables]
+        return [self.outcome_variable.name]
```
jmafoster1, line 60: Does this really need to be a list if it's only ever going to contain one element? I know we want to keep in mind that we eventually want to support multiple treatments but this seems a little illogical...

causal_testing/testing/causal_test_case.py
```diff
@@ -69,7 +70,8 @@ def get_treatment_values(self):
     def __str__(self):
         treatment_config = {k.name: v for k, v in self.treatment_input_configuration.items()}
         control_config = {k.name: v for k, v in self.control_input_configuration.items()}
+        outcome_variable = {self.outcome_variable}
```
jmafoster1, line 73: Same here, for the single element set

causal_testing/testing/causal_test_case.py
cwild-UoS: Typo, thanks for spotting

causal_testing/testing/causal_test_case.py
cwild-UoS: I completely agree, I started refactoring this and I think it will make the code in many places much neater removing the dictionary configuartions. However looking at it the scope of it definitely deserves it's own PR so i've moved it onto a separate branch and will make a new PR in time :)

causal_testing/specification/causal_dag.py
cwild-UoS: The treatment and outcome `Variable`s are also stored in the `base_test_case` and are used in the identification function. 

I designed the `base_test_case` to hopefully store exactly what is required for identification

jmafoster1: Ah, that's sensible then

causal_testing/json_front/json_class.py
cwild-UoS: test["mutations"] returns a dictionary with a single key:value pair in it. Because it's unordered `[0]` doesn't work so the extra `next(iter( ))` is the cleanest way I can think of. I'm not a big fan of it either however, I'll also think if there's other ways!

causal_testing/generation/abstract_causal_test_case.py
```diff
@@ -29,7 +30,7 @@ def __init__(
         effect_modifiers: set[Variable] = None,
         estimate_type: str = "ate",
     ):
-        assert treatment_variables.issubset(scenario.variables.values()), (
+        assert {treatment_variables}.issubset(scenario.variables.values()), (
```
jmafoster1, line 33: Could we not just do `in` instead of making a singleton set?

jmafoster1: Also we could maybe call it `treatment_variable` rather than `treatment_variables` plural if there's only one...

causal_testing/json_front/json_class.py
jmafoster1: Thinking about it, we could probably insist on a single mutation which is the treatment variable. I made it a dict when we were going to support multiple mutations.

causal_testing/generation/abstract_causal_test_case.py
cwild-UoS: Thanks Michael, that's a lot nicer implementation.

cwild-UoS: Also I've got another branch going to clean up all the multiple variable refactors so it will done in there :)

causal_testing/json_front/json_class.py
cwild-UoS: Next PR will address this

tests/testing_tests/test_causal_test_suite.py
```diff
@@ -0,0 +1,132 @@
...
+        f = open(dag_dot_path, "w")
+        f.write(dag_dot)
+        f.close()
+
+        np.random.seed(1)
+        df = pd.DataFrame({"D": list(np.random.normal(60, 10, 1000))})  # D = exogenous
+        df["A"] = [1 if d > 50 else 0 for d in df["D"]]
+        df["C"] = df["D"] + (4 * (df["A"] + 2))  # C = (4*(A+2)) + D
+        self.observational_data_csv_path = os.path.join(temp_dir_path, "observational_data.csv")
+        df.to_csv(self.observational_data_csv_path, index=False)
```
bobturneruk, line 45: This is standard across our tests, to create test data dynamically, rather than cache it somewhere?

tests/testing_tests/test_causal_test_suite.py
```diff
@@ -0,0 +1,132 @@
...
+
+        self.assertEqual(test_case, manual_test_case)
+
+    def test_execute_test_suite_single_base_test_case(self):
+        """Check that the test suite can return the correct results from dummy data for a single base_test-case"""
+        causal_test_engine = self.create_causal_test_engine()
+
+        causal_test_results = causal_test_engine.execute_test_suite(test_suite=self.test_suite)
+        causal_test_case_result = causal_test_results[self.base_test_case]
+        self.assertAlmostEqual(causal_test_case_result["LinearRegressionEstimator"][0].test_value.value, 4, delta=1e-10)
```
bobturneruk, line 103: How do we know the answer should be 4? i.e. is this a regression test - it was 4 last time we run it, or do we have an independent calculation?

bobturneruk (approved):
Looks good. Some suggestions. My usual caveats around not being super familiar with the totality of the code resulting in my comments mostly about lines of code in isolation.

causal_testing/testing/causal_test_case.py
```diff
@@ -36,12 +36,13 @@ def __init__(
         :param treatment_input_configuration: The input configuration representing the treatment values of the treatment
```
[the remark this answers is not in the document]
cwild-UoS: Definitely, I should have included it in this PR but it is instead all in #118.

causal_testing/testing/causal_test_suite.py
```diff
@@ -0,0 +1,27 @@
...
+    functionality of this class.
+    The dictionary structure should be the keys are base_test_cases representing the treatment and outcome Variables,
+    and the values are test objects. Test Objects hold a causal_test_case_list which is a list of causal_test_cases
+    which provide control and treatment values, and an iterator of Estimator Class References
+
+    This dictionary can be fed into the CausalTestEngines execute_test_suite function which will iterate over all the
+    base_test_case's and execute each causal_test_case with each iterator.
+    """
+
+    def add_test_object(self, base_test_case, causal_test_case_list, estimators, estimate_type: str = "ate"):
```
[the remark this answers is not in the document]
cwild-UoS: It would be nice to have that resilience for non intentional stuff. I'm not sure how a multi-key `__setitem__` would work for a dictionary subclass without explicitly calling the `__setitem__` method however?

examples/lr91/causal_test_max_conductances_test_suite.py
```diff
@@ -0,0 +1,190 @@
+import pandas as pd
```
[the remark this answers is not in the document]
cwild-UoS: Ahh, thanks for pointing this out, it does need to be documented better. 

`causal_test_max_conductances.py` is the original example which has only been updated to function with the new `base_test_case` functionality.  

`causal_test_max_conductances_test_suite.py` makes use of the new `test_suite` functionality rather than using custom loops like the original. 

I think having both for comparison may be useful to show why the test_suite is more concise? I'll add a readme explaining the differences!

tests/testing_tests/test_causal_test_suite.py
cwild-UoS: It is how the current `causal_test_engine` tests create / use data:

https://github.com/CITCOM-project/CausalTestingFramework/blob/main/tests/testing_tests/test_causal_test_engine.py.

I've not played with caching test data, but it makes sense and I'll make an issue to be looked into

tests/testing_tests/test_causal_test_suite.py
cwild-UoS: This test is a reimplementation of an existing `test_causal_test_engine` test case using the `causal_test_suite`:
https://github.com/CITCOM-project/CausalTestingFramework/blob/8d77deadc5e7ff819d0356fa1292e4cea07dbc59/tests/testing_tests/test_causal_test_engine.py#L159

So the source data and the result are identical to this test, so I'm not sure why 4 is the answer.

(Also feels a bit like a hitchhikers guide reference this)

tests/testing_tests/test_causal_test_suite.py
```diff
@@ -0,0 +1,132 @@
+import unittest
+import os
+import numpy as np
+import pandas as pd
+from causal_testing.testing.causal_test_engine import CausalTestEngine
```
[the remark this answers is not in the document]
cwild-UoS: https://peps.python.org/pep-0008/#imports

It looks like imports should usually be on separated lines.

causal_testing/testing/causal_test_suite.py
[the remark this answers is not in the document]
bobturneruk: Yeah - you're right. Forget this suggestion!

tests/testing_tests/test_causal_test_suite.py
bobturneruk: Fair enough - this make sense @jmafoster1, @AndrewC19 ?

AndrewC19: That makes sense. We synthesise the data on line 55 (`df["C"] = df["D"] + (4 * (df["A"] + 2))  # C = (4*(A+2)) + D`) such that the expected effect is 4.

causal_testing/specification/causal_dag.py
```diff
@@ -465,5 +465,28 @@ def depends_on_outputs(self, node: Node, scenario: Scenario) -> bool:
...
+        if base_test_case.effect == "total":
+            minimal_adjustment_sets = self.enumerate_minimal_adjustment_sets(
+                [base_test_case.treatment_variable.name], [base_test_case.outcome_variable.name]
+            )
+        elif base_test_case.effect == "direct":
+            minimal_adjustment_sets = self.direct_effect_adjustment_sets(
+                [base_test_case.treatment_variable.name], [base_test_case.outcome_variable.name]
+            )
+        else:
+            raise ValueError("Causal effect should be 'total' or 'direct'")
```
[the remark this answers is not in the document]
bobturneruk: Couple of options to explore if you're interested, neither necessarily "better" than what you're doing already...

```
class effect(str):
    "A subclass of str that can only be 'total' or 'direct'."

    def __new__(cls, string):
        if not string in ["total", "direct"]:
            raise ValueError(f"Invalid effect: {string}")

        super().__new__(cls, string)
```

```
from enum import Enum


class effect(Enum):
    "An enumeration of allowable effect types."
    total = "total"
    direct = "direct"
```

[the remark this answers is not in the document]
jmafoster1: Personally, I'm a bit fan of enums now. I use them a lot when I'm doing functional programming, and I'm guessing we can do case switches on them in Python 3.10+

causal_testing/testing/causal_test_engine.py
```diff
@@ -44,39 +46,69 @@ class CausalTestEngine:
...
-        treatment_variables = list(causal_test_case.control_input_configuration)
-        if causal_test_case.effect == "total":
-            minimal_adjustment_sets = self.casual_dag.enumerate_minimal_adjustment_sets(
-                [v.name for v in treatment_variables], [v.name for v in causal_test_case.outcome_variables]
-            )
-        elif causal_test_case.effect == "direct":
-            minimal_adjustment_sets = self.casual_dag.direct_effect_adjustment_sets(
-                [v.name for v in treatment_variables], [v.name for v in causal_test_case.outcome_variables]
+        if self.scenario_execution_data_df.empty:
+            raise Exception("No data has been loaded. Please call load_data prior to executing a causal test case.")
```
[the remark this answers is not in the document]
cwild-UoS: good idea! Added this thanks bob :)

causal_testing/specification/causal_dag.py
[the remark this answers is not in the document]
cwild-UoS: I like enums as well which I've now added

causal_testing/testing/causal_test_engine.py
```diff
@@ -44,39 +46,69 @@ class CausalTestEngine:
...
+        test_suite_results = {}
+        for edge in test_suite:
+            print("edge: ")
```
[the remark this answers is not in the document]
cwild-UoS: The logging is only setup to run from the JSON frontend right now. 

It does make sense to refactor this throughout the repo though so I'll raise an issue to do this

causal_testing/testing/causal_test_engine.py
```diff
@@ -44,39 +46,69 @@ class CausalTestEngine:
...
+
+            estimators = test_suite[edge]["estimators"]
+            tests = test_suite[edge]["tests"]
+            estimate_type = test_suite[edge]["estimate_type"]
+            results = {}
+            for EstimatorClass in estimators:
+                causal_test_results = []
+
+                for test in tests:
+                    treatment_variable = list(test.treatment_input_configuration.keys())[0]
```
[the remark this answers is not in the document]
cwild-UoS: This is what is resolved in #118! Wasn't a fan of this variable structure either

causal_testing/testing/causal_test_engine.py
```diff
@@ -44,39 +46,69 @@ class CausalTestEngine:
...
             causal_specification.causal_dag,
             causal_specification.scenario,
         )
         self.data_collector = data_collector
         self.scenario_execution_data_df = self.data_collector.collect_data(**kwargs)
-        self.minimal_adjustment_set = set()
 
-    def identification(self, causal_test_case):
-        """Identify and return the minimum adjustment set
+    def execute_test_suite(self, test_suite: CausalTestSuite) -> list[CausalTestResult]:
```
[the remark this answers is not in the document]
cwild-UoS: Exactly :). Both are options! Before to execute multiple tests you executed single tests in a loop. Which can now be instead done from 1 call with a suite.

causal_testing/testing/causal_test_engine.py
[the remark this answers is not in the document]
bobturneruk: Yeah - we really don't want `print` statements in there. Maybe there needs to be two logs somehow given how the log is the main output for the json side of things. I think a different output format for the actual output might ultimately be even better.

PR #108, opened 2022-11-29
z3 dependency fix

### LINKED ISSUE #107: Json Front End(?) causing issue with out of date dependency
I have encountered the bug outlined in this issue on the z3 library for every run of the json front end: https://github.com/Z3Prover/z3/issues/5656
This occurs on the python teardown after the tests have concluded, independent of the tests.

We are running z3 version `z3_solver~=4.8.13.0`. This bug has been fixed in version 4.11.2 (https://github.com/Z3Prover/z3/commit/e4683863592f5118c6ccdc4164bcdccd3b1369de)

Should we update this dependency?

rsomers1998:
closes #107 

Updated dependency to fix bug related to https://github.com/Z3Prover/z3/issues/5656

Framework no longer throws errors on shutting down z3 variables

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,7 +11,7 @@ requirements = [
     "scipy~=1.7.3",
     "statsmodels~=0.13.2",
     "tabulate~=0.8.10",
-    "z3_solver>=4.11.2",
+    "z3_solver~=4.11.2",
 ]
 
 # Additional dependencies for development
--- a/setup.py
+++ b/setup.py
@@ -11,7 +11,7 @@ requirements = [
     "scipy~=1.7.3",
     "statsmodels~=0.13.2",
     "tabulate~=0.8.10",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -11,7 +11,7 @@
     "scipy~=1.7.3",
     "statsmodels~=0.13.2",
     "tabulate~=0.8.10",
-    "z3_solver~=4.8.13.0",
+    "z3_solver>=4.11.2",
```
cwild-UoS, line 14: Is there a reason to swap from the `Compatible release clause` `~=` to the greater or equal to clause `>=` ?

cwild-UoS (approved):

setup.py
rsomers1998: I didn't 100% understand what `~=` meant. After a quick google, I'll swap it to that.

PR #109, opened 2022-11-30
Metamorphic relation and test generation from a DAG

AndrewC19:
This PR adds functionality for generating metamorphic relations from a causal DAG and support for generating metamorphic test cases for these metamorphic relations, if the variables involved are numerical.


jmafoster1 (commented):
I'm a little worried that some of the lines in `MetamorphicRelation` are uncovered. I'm not a fan of writing tests just for the sake of it, but stuff like line 61 should probably be tested.
Also, @bobturneruk @cwild-UoS is there a way to get the abstract methods to pick up as covered if their concrete implementations are covered? We obviously can't test the pure abstract methods...

bobturneruk:
@jmafoster1 - is the query above about testing ABCs still open? I've had a bit of a dig, and I'll come back with some info. if it's useful...

jmafoster1:
@bobturneruk yes still an open question. I'm interested to hear what you've found.

bobturneruk:
Would removing the ellipsis (not sure what the plural is) solve this adequately? I'm probably being too superficial.

bobturneruk:
Probably doesn't need the ellipsis or `pass`.

jmafoster1:
Will it not throw horrible syntax errors if you just leave the method entirely blank?

jmafoster1 (approved):

PR #113, opened 2022-12-19
Robustness Estimators

rsomers1998:
A draft PR to keep track of changes made when implementing robustness checks against a causal inference claim.


rsomers1998:
closes #167

causal_testing/testing/validation.py
```diff
@@ -0,0 +1,56 @@
...
+        """Calculate the E value from a risk ratio. This allow
+        the user to identify how large a risk an unidentified confounding
+        variable would need to be to nullify the causal relationship
+        under test."""
+
+        if risk_ratio >= 1:
+            e = risk_ratio + math.sqrt(risk_ratio * (risk_ratio - 1))
+
+            lower_limit = confidence_intervals[0]
+            if lower_limit <= 1:
```
jmafoster1, line 39: Should lower and upper limit both be testing for 1 in this way? Purely guessing from the variable names, I'd have said that we should have `lower_limit <= something <= upper_limit`, but with these tests, we'd end up with lower limit being greater than upper limit, or them both being set to 1.

rsomers1998: Mis implementation on my part. lower and upper limits shouldn't be returned or changed, they should only affect the e value if present. I have reimplemented this as both a with and without ci's method. Table 1 of https://oce.ovid.com/article/00000605-201708150-00007/PDF has been followed for the current implementation

causal_testing/testing/validation.py
```diff
@@ -0,0 +1,59 @@
...
+                e = lower_limit + math.sqrt(lower_limit * (lower_limit - 1))
+
+            return e
+
+        upper_limit = confidence_intervals[1]
+        e = 1
+        if upper_limit < 1:
+            upper_limit_prime = 1 / upper_limit
+            e = upper_limit_prime + math.sqrt(upper_limit_prime * (upper_limit_prime - 1))
+
```
jmafoster1, line 58: Similar here for `If UL ≥ 1, then E-value = 1`

causal_testing/testing/validation.py
```diff
@@ -0,0 +1,59 @@
...
+        This allow the user to identify how large a risk an unidentified
+        confounding variable would need to be to nullify the causal relationship
+        under test."""
+
+        if risk_ratio >= 1:
+            lower_limit = confidence_intervals[0]
+            e = 1
+            if lower_limit > 1:
+                e = lower_limit + math.sqrt(lower_limit * (lower_limit - 1))
+
```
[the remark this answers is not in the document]
rsomers1998: Line 47 handles this, e is set to 1 and never changed in 49 if LL <= 1. It felt cleaner with variable scopes to do it this way.

causal_testing/testing/validation.py
rsomers1998: See above, but for line 54

causal_testing/testing/validation.py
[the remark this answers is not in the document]
jmafoster1: Ah course it does! Nice one

jmafoster1 (approved):

PR #116, opened 2022-12-22
Added confidence intervals to the logistic regression estimator

jmafoster1:
I added confidence intervals to the logistic regression estimator since the CTF seems to expect them. No rush for reviews

### DIFF
--- a/causal_testing/testing/estimators.py
+++ b/causal_testing/testing/estimators.py
@@ -149,7 +149,8 @@ class LogisticRegressionEstimator(Estimator):
     def estimate_control_treatment(self) -> tuple[pd.Series, pd.Series]:
         """Estimate the outcomes under control and treatment.
 
-        :return: The average treatment effect and the 95% Wald confidence intervals.
+        :return: The estimated control and treatment values and their confidence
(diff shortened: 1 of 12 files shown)


causal_testing/testing/estimators.py
```diff
@@ -168,31 +169,53 @@ def estimate_control_treatment(self) -> tuple[pd.Series, pd.Series]:
...
 
         y = model.predict(x)
-        return y.iloc[1], y.iloc[0]
+
+        # Delta method confidence intervals from
+        # https://stackoverflow.com/questions/47414842/confidence-interval-of-probability-prediction-from-logistic-regression-statsmode
+        cov = model.cov_params()
+        gradient = (y * (1 - y) * x.T).T  # matrix of gradients for each observation
+        std_errors = np.array([np.sqrt(np.dot(np.dot(g, cov), g)) for g in gradient.to_numpy()])
+        c = 1.96  # multiplier for confidence interval
```
cwild-UoS, line 178: Just double checking that this is a constant that will never change and therefore okay to hardcode?

cwild-UoS (approved):
Looks good to me!

PR #119, opened 2023-01-04
causal_test_result + new dataclass for test values

rsomers1998:
CausalTestResult moved from causal test outcome to causal test result
TestValue dataclass added to hold different test result values and their types
CausalTestOutcomes updated so `apply` method can take different test values

I have no clue what else this breaks (example code etc)

### DIFF
--- a/causal_testing/testing/causal_test_outcome.py
+++ b/causal_testing/testing/causal_test_outcome.py
@@ -1,92 +1,6 @@
 from abc import ABC, abstractmethod
-from typing import Any, Union
-
+from causal_test_result import CausalTestResult
 import numpy as np
-
-from causal_testing.specification.variable import Variable
-
-
-class CausalTestResult:
-    """A container to hold the results of a causal test case. Every causal test case provides a point estimate of
-    the ATE, given a particular treatment, outcome, and adjustment set. Some but not all estimators can provide
-    confidence intervals."""
-
-    def __init__(
-        self,
-        treatment: tuple,
-        outcome: tuple,
-        treatment_value: Union[int, float, str],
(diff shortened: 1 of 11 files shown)


causal_testing/testing/causal_test_outcome.py
```diff
@@ -1,93 +1,7 @@
 from abc import ABC, abstractmethod
-from typing import Any, Union
-
+from causal_test_result import CausalTestResult
```
AndrewC19, line 2: This import needs to be changed to `from causal_testing.testing.causal_test_result import CausalTestResult`.

rsomers1998: Fixed

cwild-UoS (commented):
There is a `black` formatting issue raised which you can fix by running `black causal_testing` in the root of the project (you may have to pip install black first).

Also the tests are failing and I suspect that it is because this import needs updating:
https://github.com/CITCOM-project/CausalTestingFramework/blob/a212722a0d58841ddc4308a61f37e0512a5e337d/tests/testing_tests/test_causal_test_outcome.py#L2

Looks all good to me otherwise :)

cwild-UoS:
Just pushed up a change to the tests, they needed to be adapted to use the new TestValue

cwild-UoS:
Just run the tests locally and there are a few more that need updating, I'll have a look at them now and push another commit with they updated tests

rsomers1998:
Just saw your changes. Hopefully, I didn't break them. Please do check though. I think I got all the tests working and the example code updated. There was a lot more than I first thought.

cwild-UoS (approved):

PR #120, opened 2023-01-13
install instructions for specific versions

bobturneruk:
Draft 1.0.0 release here https://github.com/CITCOM-project/CausalTestingFramework/releases/edit/untagged-09a6985599a8f8f7aa2d should generate `zip` and `tar.gz` on publish. We would still expect people to install from github clone rather than these files. Instructions for this are in this PR. They're too clunky so we need to get on PyPI soon.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Here are some explanations for the causal inference terminology used above.
 
 ## Installation
 
-To use the causal testing framework, clone the repository, `cd` into the root directory, and run `pip install -e .`. More detailled installation instructions can be found in the [online documentation](https://causal-testing-framework.readthedocs.io/en/latest/installation.html).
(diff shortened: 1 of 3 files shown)


cwild-UoS (approved):
Looks good, I think the install instructions are really clear.

Having looked at the semantic version guide:
https://semver.org/spec/v2.0.0.html
It could be worth changing the release version depending on how we view this release.

Is this a pre-release which deserves an alpha/beta tag?

Or do we treat it as the initial development release and give it the tag v0.1.0?
https://semver.org/spec/v2.0.0.html#how-should-i-deal-with-revisions-in-the-0yz-initial-development-phase

Not sure which is the best choice or if it is even an impactful one!

bobturneruk:
> If you’re worrying a lot about backwards compatibility, you should probably already be 1.0.0.

I am 😄

jmafoster1 (approved):

AndrewC19 (approved):

PR #132, opened 2023-02-01
Add github actions to publish to PyPI

cwild-UoS:
There are two pipelines, one to publish to [test PyPI](https://test.pypi.org/) upon a pull request merge to main and to [PyPI ](https://pypi.org/) on a new release. 

Can't test these pipelines until the next pull request as far as I'm aware


cwild-UoS:
The main `README.md` has become out of sync with the latest changes in CTF. So I will make a PR with the updated `README.md` which will also serve as a good test for this?

bobturneruk (changes requested):
Good progress. I think you should just put the test one in for now, and look at how you're triggering things. I suggest you want this:

```
on:
  push:
    tags:
      - v*
```

To trigger on tagged releases. I guess your readme PR gives you a patch version to test on.

cwild-UoS:
Yes the README PR will hopefully allow me to test the different triggers, should have been more clear!

I planned on having the test action trigger on PRs as it can serve as a test for the build as well as creating a potential place where you can share a non-released version.

bobturneruk:
I guess that might mean you push the same version twice to testpypi - will it allow this?

bobturneruk:
Nope. That might work just as well. One for #team-python, perhaps.

cwild-UoS:
> I guess that might mean you push the same version twice to testpypi - will it allow this?

The guide given by [python.org](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/) shows an example action that:
`These two steps use the [pypa/gh-action-pypi-publish](https://github.com/pypa/gh-action-pypi-publish) GitHub Action: the first one uploads contents of the dist/ folder into TestPyPI unconditionally and the second does that to PyPI, but only if the current commit is tagged.`

Which is where I got the idea but good point! I can't find anywhere that states you can do that for testpypi

cwild-UoS:
Tying `z3_solver` to `z3_solver~=4.11` installs version `4.12.1.0` which is not compatible with `4.11.2.0` which is what it was previously tied to.

It looks like z3 does not comply with semantic versioning so I've tied it back to the `4.11.2` for now

bobturneruk (approved):
Looks good! Let's see what happens when you merge and give it a test!

PR #135, opened 2023-02-09
Linting 

cwild-UoS:


cwild-UoS:
Pylint codes to maybe discuss:
`C0114: Missing module docstring (missing-module-docstring)`
`R0903: Too few public methods (1/2) (too-few-public-methods)`

cwild-UoS:
Remaining linting issues are:
`R0902: Too many instance attributes (8/7)`
`R0903: Too few public methods`
`R0913: Too many arguments`
`R0914: Too many local variables`
`C0114: Missing module docstring`

`R0902`, `R0903` and `R0914` I'll do a follow up PR to resolve so this one doesn't get too large and as they will require some refactoring to break up the classes that are doing too much.

cwild-UoS:
In regards to `C0114: Missing module docstring`, do you think @bobturneruk module docstrings will be beneficial to the project? I'm not sure as a lot of the modules in the project contain a single class, so the module docstring feels unnecessary

bobturneruk:
I would suggest a short docstring in all modules - I guess because in some modules it will be really useful. Or you could disable it on a per module basis.

cwild-UoS:
`R0903: Too few public methods` 

One other question is about this linting warning. The two warnings we have for it, are extensions of the CausalTestOutcome class here: https://github.com/CITCOM-project/CausalTestingFramework/blob/1098043c96a0d7355ae7e60c6e1b1805822f56b5/causal_testing/testing/causal_test_outcome.py#L37

This has been raised as an issue on pylint, where the are multiple public methods in the parent, but the child only has 1 public method (pylint wants 2 public methods per class by default). It would seem sensible to maybe drop the pylint public method expectation to 1? 

https://github.com/PyCQA/pylint/issues/4352

bobturneruk:
Could disable for specific files?

causal_testing/specification/variable.py
```diff
@@ -161,15 +165,14 @@ def cast(self, val: Any) -> T:
             return float(val.numerator().as_long() / val.denominator().as_long())
         if hasattr(val, "is_string_value") and val.is_string_value() and self.datatype == str:
             return val.as_string()
-        if (isinstance(val, float) or isinstance(val, int) or isinstance(val, bool)) and (
-            self.datatype == int or self.datatype == float or self.datatype == bool
-        ):
+        if (isinstance(val, (float, int, bool))) and (self.datatype in (float, int, bool)):
```
bobturneruk, line 168: Nice and neat.

causal_testing/testing/causal_test_outcome.py
```diff
@@ -35,22 +43,22 @@ class Positive(CausalTestOutcome):
...
     """An extension of TestOutcome representing that the expected causal effect should be negative."""
 
     def apply(self, res: CausalTestResult) -> bool:
-        # TODO: confidence intervals?
         if res.test_value.type == "ate":
             return res.test_value.value < 0
-        elif res.test_value.type == "risk_ratio":
+        if res.test_value.type == "risk_ratio":
             return res.test_value.value < 1
+        raise ValueError(f"Test Value type {res.test_value.type} is not valid for this TestOutcome")
```
bobturneruk, line 61: One can test that correct errors are thrown in the right circumstance.

causal_testing/testing/effect.py
```diff
@@ -1,7 +1,10 @@
...
 from enum import Enum
 
 
 class Effect(Enum):
-    "An enumeration of allowable effect types."
-    total = "total"
-    direct = "direct"
+    """An enumeration of allowable effect types."""
+
+    TOTAL = "total"
```
bobturneruk, line 9: Guess this is the case CONVENTION!

bobturneruk (approved):
Looking good. A few questions / suggestions.

causal_testing/json_front/json_class.py
```diff
@@ -218,7 +221,7 @@ def _setup_test(self, causal_test_case: CausalTestCase, estimator: Estimator) ->
 
         return causal_test_engine, estimation_model
 
-    def add_modelling_assumptions(self, estimation_model: Estimator):
+    def add_modelling_assumptions(self, estimation_model: Estimator):  # pylint: disable=unused-argument
```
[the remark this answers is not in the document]
cwild-UoS: I think you're right that it doesn't make much sense in it's current implementation.

It isn't the same method but is still related to `estimators.py`, I've made a new [issue ](https://github.com/CITCOM-project/CausalTestingFramework/issues/140) to fix and better document what this method does. It's an optional method to expose some of the functionality of estimator objects, there is an implementation in the `Poisson` example [here](https://github.com/CITCOM-project/CausalTestingFramework/blob/e9f51e5d01ba6f5abc0cf9093f8c607ea4cee996/examples/poisson/run_causal_tests.py#L150)

causal_testing/generation/abstract_causal_test_case.py
```diff
@@ -256,3 +243,30 @@ def generate_concrete_tests(
                 len(concrete_tests),
             )
         return concrete_tests, runs
+
+    def _optimizer_model(self, run_columns: Iterable[str], row: pd.core.series) -> z3.Optimize:
```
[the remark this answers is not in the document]
cwild-UoS: Yes, the [_generate_concrete_tests](https://github.com/CITCOM-project/CausalTestingFramework/blob/e9f51e5d01ba6f5abc0cf9093f8c607ea4cee996/causal_testing/generation/abstract_causal_test_case.py#L74) method had a `too-many-locals` linting error. The `_optimizer_model` function takes the setup of the optimizer model out of the original method, into it's own function to reduce how much the original method is doing and more logically breakup the functionality. Resolving the `too-many-locals` linting warning.

.pylintrc
```diff
@@ -152,6 +152,7 @@ disable=raw-checker-failed,
         useless-suppression,
         deprecated-pragma,
         use-symbolic-message-instead,
+        logging-fstring-interpolation,
```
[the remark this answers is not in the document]
cwild-UoS: Yeah we had a discussion about this quite a while ago IIRC. There are some logging strings that Micahel and I have implemented using the %-formatting that are quite long and messy. 

It does allow for the lazily evaluation of the logging strings, but I think the readability is nicer than the very marginal performance differences.

causal_testing/testing/causal_test_outcome.py
cwild-UoS: Sounds worth doing 👍. I'll save it for another PR to keep this just linting

PR #139, opened 2023-02-14
Product terms now accessible in LR estimator

AndrewC19:
In this PR, I have made a minor change to the init method of the `LinearRegressionEstimator` class to ensure that `self.product_terms` is created before `self.add_product_term_to_df` is called.

Previously the initialiser was calling the method `self.add_product_term_to_df`(which adds pairs of variables to the `self.product_terms` list) before `self.product_terms` is created.

I have also added a test for a simple causal effect with an interaction. This test failed due to this bug and now passes after the fix.

### DIFF
--- a/causal_testing/testing/estimators.py
+++ b/causal_testing/testing/estimators.py
@@ -300,7 +300,7 @@ class LinearRegressionEstimator(Estimator):
         self.intercept = intercept
 
         if product_terms:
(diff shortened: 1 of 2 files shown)


jmafoster1 (approved):
I plan to start looking at refactoring all this this afternoon actually so the user is able to supply a functional form rather than having to specify all these extra terms, but I'll approve for now

PR #141, opened 2023-02-20
Fix PyPI github action

cwild-UoS:
The action failed due to 
```
ERROR Missing dependencies:
	setuptools_scm[toml]>=6.2
Error: Process completed with exit code 1.
```
So I've added an instruction to install `setuptools_scm` and tested it on a new local environment.


bobturneruk (approved):
Makes sense!

PR #142, opened 2023-02-20
Pylint refactoring

cwild-UoS:
PR for the remaining Pylint issues that will involve some level of refactoring of current classes and methods


cwild-UoS:
Disabled the `import-error` check due to the need for megalinter to run in an environment with all the python modules installed. 

It is possible to install python packages to the megalinter environment using [pre-commands](https://megalinter.io/latest/configuration/#pre-commands). However it seems very time consuming and doesn't sync well with local development?

jmafoster1 (approved):

PR #148, opened 2023-02-24
Py pi action setuptool version

cwild-UoS:
Currently the `Publish python PyPI test` action is hitting the error:

```
ERROR Missing dependencies:
	setuptools>=64
```

The version installed is 
`Requirement already satisfied: setuptools in /opt/hostedtoolcache/Python/3.9.16/x64/lib/python3.9/site-packages (from numba->shap<0.41.0,>=0.38.1->econml~=0.12->causal-testing-framework==2.0.3) (58.1.0)`

### DIFF
--- a/.github/workflows/publish-to-pypi.yaml
+++ b/.github/workflows/publish-to-pypi.yaml
@@ -1,5 +1,9 @@
 name: Publish python PyPI
 
+on:
+  push:
+    tags:
+      - v0.0.0
 jobs:
   build-release:
     name: Build and publish PyPI
--- a/.github/workflows/publish-to-test-pypi.yaml
+++ b/.github/workflows/publish-to-test-pypi.yaml
@@ -22,6 +22,7 @@ jobs:
           pip3 install .
           pip3 install .[pypi]
           pip3 install build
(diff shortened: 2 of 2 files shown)


cwild-UoS:
Is there a nicer way of handling this do you know @bobturneruk?

bobturneruk:
Ensuring the required version of setuptools is on the runner is the nicest way I can think of.

.github/workflows/publish-to-pypi.yaml
```diff
@@ -1,5 +1,9 @@
 name: Publish python PyPI
 
+on:
+  push:
+    tags:
+      - v0.0.0
```
bobturneruk, line 6: This will change?

bobturneruk (approved):
Sensible plan re: setuptools.

.github/workflows/publish-to-pypi.yaml
cwild-UoS: Yes! Should have put a comment next to it. 

It's only temporary because without a trigger it is causing a lot of erroneous actions to be ran

PR #150, opened 2023-03-03
Publish to PyPI

cwild-UoS:


cwild-UoS:
Apologies for being slow with this! 

I was having some issues with installing from test.pypi.org. However I realised that it was trying to resolve dependencies from test.pypi.org as well which doesn't have the latest packages.

So I believe it is all ready for real PyPI :)

bobturneruk (approved):
Looks good, let's see what happens!

PR #154, opened 2023-03-09
Silenced logistic regression fitting output

jmafoster1:
Small PR to turn off the default logging behaviour when fitting logistic regression models as it's annoying.

### DIFF
--- a/causal_testing/testing/estimators.py
+++ b/causal_testing/testing/estimators.py
@@ -152,8 +152,7 @@ class LogisticRegressionEstimator(Estimator):
                 treatment_and_adjustments_cols = pd.get_dummies(
                     treatment_and_adjustments_cols, columns=[col], drop_first=True
                 )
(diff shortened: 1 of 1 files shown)


causal_testing/testing/estimators.py
```diff
@@ -152,8 +152,7 @@ def _run_logistic_regression(self, data) -> RegressionResultsWrapper:
                 treatment_and_adjustments_cols = pd.get_dummies(
                     treatment_and_adjustments_cols, columns=[col], drop_first=True
                 )
-        regression = sm.Logit(outcome_col, treatment_and_adjustments_cols)
-        model = regression.fit()
```
cwild-UoS, line 156: Is it intentional to remove calling the .fit() function?

jmafoster1: I didn't remove it. I just made it a one liner since the only use of `regression` was to call `regression.fit()`, so I merged the two lines into one

cwild-UoS (approved):

PR #162, opened 2023-03-13
Json read multiple files

cwild-UoS:
Allows for multiple CSVs to be specified with the Json Frontend, which will concatenate them assuming the same columns are in each CSV. 

`python foo.py --data_path "data/bar1.csv" "data/bar2.csv" --dag_path="data/bar.dot" --json_path="data/bar.json"

### DIFF
--- a/causal_testing/data_collection/data_collector.py
+++ b/causal_testing/data_collection/data_collector.py
@@ -124,9 +124,9 @@ class ExperimentalDataCollector(DataCollector):
 class ObservationalDataCollector(DataCollector):
     """A data collector that extracts data that is relevant to the specified scenario from a csv of execution data."""
 
-    def __init__(self, scenario: Scenario, csv_path: str):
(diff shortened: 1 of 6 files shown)


cwild-UoS:
However, this introduces a breaking change to the Observational. Where instead of the path of a `.csv` containing the dataframe is passed to the class, just the dataframe is passed.

https://github.com/CITCOM-project/CausalTestingFramework/blob/18b5716579f2181b7980e4a0063df7b2e9a1b6f3/causal_testing/data_collection/data_collector.py#L127

Which means any code using the `ObservatonalDataCollector` class will need to be refactored.

jmafoster1:
Codecov seems very unhappy with this? `Codecov: Failed to properly upload: The process '/home/runner/work/_actions/codecov/codecov-action/v2/dist/codecov' failed with exit code 255`

bobturneruk:
codecove mood has improved it seems.

jmafoster1 (approved):

PR #163, opened 2023-03-13
Functional form

jmafoster1:
We can now specify regression equations as strings rather than having to use `add_squared_term` and all the rest


AndrewC19 (approved):

cwild-UoS (approved):
Looks great to me!

PR #164, opened 2023-03-16
Examples now runnable with pytest. Closes #5. Closes #136.

### LINKED ISSUE #5: Add testing guidelines
Within the tests directory, we should add a set of testing guidelines that give a brief introduction into the testing practices and tools used in the causal testing framework.

Points to cover:
- [x] Unit Tests: Each method and class should be well-covered by a set of unit tests.
- [x] Regression Tests: Upon PR, it should be confirmed that no previously passing test cases now fail.
- [ ] Integration Tests: When multiple parts of the casual testing framework are implemented, we should create a suite of integration tests which confirm they work together correctly e.g. identification and estimation.
- [x] Code Coverage: State what we consider to be an acceptable level of code coverage e.g. that each PR must not reduce coverage.

### LINKED ISSUE #136: Regression test examples
With large refactors of the framework happening, would be nice to regression test examples to see if they are stable

jmafoster1:


jmafoster1:
Tests are failing because of missing `covasim` dependency. I don't want to make this a core dependency of the project. Is there a way to just make it a dependency for running the tests on the Github action?

cwild-UoS:
> Tests are failing because of missing `covasim` dependency. I don't want to make this a core dependency of the project. Is there a way to just make it a dependency for running the tests on the Github action?

I've added the [test] dependencies you made to the GH action workflow with `pip install -e .[test]`.  Which hopefully will do the trick

cwild-UoS (approved):

PR #171, opened 2023-03-23
Fix modelling_inputs in the poisson example. Closes #106.

### LINKED ISSUE #106: Json Front End example code bug
When defining the modelling scenario for the causal tests, the following code is run:
```
# Create input structure required to create a modelling scenario
modelling_inputs = [Input(i['name'], i['type'], i['distribution']) for i in inputs] + \
[Output(i['name'], i['type']) for i in outputs] + \
[Meta(i['name'], i['type'], [i['populate']]) for i in metas] if metas else list()
```

If metas is [] or None, the else statement does not only revert the metas variable to [], but instead changes the whole of modelling_inputs to [].
This code, with the extra brackets, works as expected but I'm sure there's a neater workaround:
```
# Create input structure required to create a modelling scenario
modelling_inputs = [Input(i['name'], i['type'], i['distribution']) for i in inputs] + \
[Output(i['name'], i['type']) for i in outputs] + \
([Meta(i['name'], i['type'], [i['populate']]) for i in metas] if metas else list())
```

cwild-UoS:


rsomers1998 (approved):
Looks good to me

PR #177, opened 2023-04-14
Json cate

jmafoster1:
Some changes to facilitate the CARLA case study. Most notably, these include rudimentary capability for CATE estimation as well as coefficient estimation, so we don't have to do concrete metamorphic tests.


causal_testing/json_front/json_class.py
```diff
@@ -188,7 +225,9 @@ def _execute_test_case(self, causal_test_case: CausalTestCase, test: Iterable[Ma
             logger.warning("   FAILED- expected %s, got %s", causal_test_case.expected_causal_effect, result_string)
         return failed
 
-    def _setup_test(self, causal_test_case: CausalTestCase, test: Mapping) -> tuple[CausalTestEngine, Estimator]:
+    def _setup_test(
+        self, causal_test_case: CausalTestCase, test: Mapping, conditions: list[str] = None
```
cwild-UoS, line 229: Might be nice to add the new `conditions` param to the docstring, but it is optional as this is a private method that users won't be interacting with. 

I believe that pylint does not pick this up as it is private

causal_testing/specification/causal_dag.py
```diff
@@ -521,5 +521,14 @@ def identification(self, base_test_case: BaseTestCase):
         minimal_adjustment_set = min(minimal_adjustment_sets, key=len)
         return minimal_adjustment_set
 
+    def to_dot(self) -> str:
```
cwild-UoS, line 524: Should this have a different name? 'to_dot' suggests you may be getting a `.dot` file back not a str?

cwild-UoS (approved):
Looks really good!

causal_testing/json_front/json_class.py
jmafoster1: Good spot. Fixed

PR #183, opened 2023-04-27
Only a single treatment now

rsomers1998:
Removed picking the first treatment when summarising results.

The current implementation just prints the first letter of the name of the treatment.


jmafoster1 (approved):
Good spot

PR #188, opened 2023-05-18
Categorical variables

jmafoster1:
Unit ATE now works for categorical variables.


causal_testing/testing/causal_test_result.py
```diff
@@ -43,18 +43,27 @@ def __init__(
...
+            result_str = "\n" + push(self.test_value.value)
         base_str = (
             f"Causal Test Result\n==============\n"
             f"Treatment: {self.estimator.treatment}\n"
             f"Control value: {self.estimator.control_value}\n"
             f"Treatment value: {self.estimator.treatment_value}\n"
             f"Outcome: {self.estimator.outcome}\n"
             f"Adjustment set: {self.adjustment_set}\n"
-            f"{self.test_value.type}: {self.test_value.value}\n"
+            f"{self.test_value.type}:{result_str}\n"
```
cwild-UoS, line 59: Very minor, would you want a space after the colon?

causal_testing/testing/causal_test_outcome.py
```diff
@@ -74,6 +85,7 @@ def apply(self, res: CausalTestResult) -> bool:
             return res.test_value.value > 0
         if res.test_value.type == "risk_ratio":
             return res.test_value.value > 1
+        # Dead code but necessary for pylint
```
cwild-UoS, line 88: Can we ensure that in all scenarios at least one of the above `if` statements will definitely run?

If a user misconfigures something then this error may still be helpful

cwild-UoS (approved):

causal_testing/json_front/json_class.py
```diff
@@ -56,13 +56,15 @@ def __init__(self, output_path: str, output_overwrite: bool = False):
         self.output_path = Path(output_path)
         self.check_file_exists(self.output_path, output_overwrite)
 
-    def set_paths(self, json_path: str, dag_path: str, data_paths: str):
+    def set_paths(self, json_path: str, dag_path: str, data_paths: list[str] = None):
```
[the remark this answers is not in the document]
jmafoster1: No. Pylint says that's "Dangerous"...

causal_testing/testing/causal_test_result.py
jmafoster1: Yes, good spot

causal_testing/testing/causal_test_outcome.py
jmafoster1: Yes, because of the call to `super` in the first `if` statement. If the user misconfigures `res.test_value.value` then this will throw a `ValueError` from there. I'm pretty sure it's impossible to reach line 89 unless we change the `super` method later on.

PR #191, opened 2023-05-22
Utils Package

cwild-UoS:
Moved `validation.py` to a new utils module. 

Also made all the methods in `validation.py` static. This helps enforce the idea that they are separated utility functions. Also the the class doesn't have to be instantiated for the methods to be used.


rsomers1998 (approved):
Looks good to me. Not 100% sure why code cov is annoyed at this change.

cwild-UoS:
> Looks good to me. Not 100% sure why code cov is annoyed at this change.

I'm a bit confused at this too, it's complaining at this line
`line 64 estimators.py`
`raise ValueError(f"Unsupported type for effect_modifiers {effect_modifiers}. Expected iterable")`

Which wasn't affected by this PR..

Issue #178, opened 2023-04-14
test_json_class side effects


jmafoster1:
**Describe the bug**
The `test_json_class.py` file creates a file `temp_out.txt` when it runs which is not cleared up after the tests run. This should be cleaned after running pytest to avoid accidentally committing it.

**To Reproduce**
1. Go to root directory
2. run `pytest tests`
3. `temp_out.txt' appears as the tests run

**Expected behavior**
Tests should clear up after themselves. `temp_out.txt` should be deleted.

**Desktop (please complete the following information):**
 - OS: Ubuntu

jmafoster1:
Along similar lines, we need to make sure we delete `json_frontend.log` files from the example directories after we run tests (or cleanup before)

cwild-UoS:
On Windows 10 interestingly `temp_out.txt` does get cleaned up, however you are right in that the code looks like it shouldn't.

jmafoster1:
Actually, I'm getting the same tests failing for the same reason for mu usual CTF environment now. The problem is that it's not our fault. It seems to be a problem with the econML package using deprecated features of numpy

jmafoster1:
FYI, here's the output of `pip freeze`
```
causal-testing-framework==4.1.1.dev12+gcf9db99
click==8.1.3
cloudpickle==2.2.1
contourpy==1.0.7
cycler==0.11.0
econml==0.14.1
fitter==1.5.2
fonttools==4.39.4
importlib-resources==5.12.0
joblib==1.2.0
kiwisolver==1.4.4
lhsmdu==1.1
lightgbm==3.3.5
llvmlite==0.40.0
matplotlib==3.7.1
networkx==2.8.8
numba==0.57.0
numpy==1.22.4
packaging==23.1
pandas==1.5.3
patsy==0.5.3
Pillow==9.5.0
pygame==2.1.2
pygraphviz==1.10
pyparsing==3.0.9
python-dateutil==2.8.2
pytz==2023.3
scikit-learn==1.2.2
scipy==1.10.1
shap==0.41.0
six==1.16.0
slicer==0.0.7
sparse==0.14.0
statsmodels==0.14.0
tabulate==0.9.0
threadpoolctl==3.1.0
tqdm==4.65.0
z3-solver==4.11.2.0
zipp==3.15.0
```

PR #194, opened 2023-06-01
Json file cleanup

christopher-wild:
Solves #178 

Could you check if this fixes the `temp_out.txt` cleanup please @jmafoster1 ? Interested to see how it runs on Ubuntu.


christopher-wild:
Another PR remove the use of the logger in the JSON frontend, and I'm not sure what the motivation to change the test temporary files like this?

jmafoster1:
Looking at the original #178, it looks like I just wanted to make sure that files created by running `pytest` didn't accidentally get committed to the repo, but if they're no longer created, then that's fine :)

PR #195, opened 2023-06-02
Can now generate metamorphic tests from the metamorphic_relation.py file

jmafoster1:
I wrote several scripts to do this in various places, so I thought I'd add the functionality into the CTF directly. It'd be cool if we could do something like `causal_testing generate_tests -d dag.dot -o tests.json` eventually, but this should do for now


cwild-UoS (commented):
Looks good to me, nice way to solve the __main__ coverage

cwild-UoS (approved):
Only left comment review last time..

PR #199, opened 2023-06-22
Move z3_solver to be first dependency

cwild-UoS:
Current PRs are failing to install z3_solver, addressing problem in this PR


cwild-UoS:
Locally installing from wheel's is fine for me, however it appears there are some remote issues with it. 

Making it an issue to investigate further, for now this will resolve the issues

jmafoster1 (approved):

PR #202, opened 2023-06-23
Estimates now taken from causal test case when executing test

rsomers1998:
`execute_test` pulled the estimate type from two places. This caused risk_ratio calculations to be performed using ATE instead. This PR fixes this

### DIFF
--- a/causal_testing/json_front/json_class.py
+++ b/causal_testing/json_front/json_class.py
@@ -229,9 +229,7 @@ class JsonUtility:
         causal_test_engine, estimation_model = self._setup_test(
             causal_test_case, test, test["conditions"] if "conditions" in test else None
         )
-        causal_test_result = causal_test_engine.execute_test(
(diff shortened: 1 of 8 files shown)


cwild-UoS:
The current megalint errors. For once they seem fairly valid. Definitely worth removing estimate_type from the docstring and updating how to JSON frontend uses this function.

If you don't fancy fighting megalint let me know :)

```
  ************* Module causal_testing.json_front.json_class
  causal_testing/json_front/json_class.py:232:29: E1123: Unexpected keyword argument 'estimate_type' in method call (unexpected-keyword-arg)
  ************* Module causal_testing.testing.causal_test_engine
  causal_testing/testing/causal_test_engine.py:99:41: E1121: Too many positional arguments for method call (too-many-function-args)
  causal_testing/testing/causal_test_engine.py:106:4: W9017: "estimate_type" differing in parameter documentation (differing-param-doc)
  ```

Issue #203, opened 2023-06-23
Risk ratio adjustment configs


rsomers1998:
The method `estimate_control_treatment` in causal_testing.testing.estimators has the argument `adjustment_config`. This variable is only exposed used in the framework and never exposed to the point where a tester can change it. This dict is required when adjusting for cate, risk ratio, etc.

Exposing this dictionary to the tester so it can added to to assign adjustment values to adjusted variables would help but doing so requires exposing it through many objects. This dictionary should be bound to a test case really as different test cases may adjust variables in different ways.

For adjustment using typical ATE, this dictionary is not needed so making it optional would be great, but also alerting the user if they have failed to specify it for required estimates types would help (eg. risk_ratio).

cwild-UoS:
Just to follow up as I'm not familiar with the execution of risk_ratios, currently the argument `adjustment_config` is optional in the method `estimate_risk_ratio`:

`def estimate_risk_ratio(self, bootstrap_size=100, adjustment_config=None) -> float:`

From what you've said above @rsomers1998, should it instead be a mandatory parameter?

cwild-UoS:
Hi Richard,

Sorry for taking a while to get around to this! Hopefully this exposes it for you, let me know if it doesn't #225

rsomers1998:
No worries at all, I've been busy with other things. This solution looks very clean and also the reimplemented estimators look more streamlined. Thanks for this.

PR #205, opened 2023-07-04
Expose estimator type params

cwild-UoS:
This PR relates #203, where parameters such as `adjustment_config` and `bootstrap_size` were not exposed to testers using the CTF. 

The way they are exposed is by a new attribute in `CausalTestCase` called `estimate_params`. Any parameters placed in here will be unpacked in the specific estimator to be used. 

e.g. `estimator_params={"adjustment_config": {"large_gauge": 0}}`

Any feedback on if this is a logical way to expose these parameters or any problems this could cause are very welcome!

### DIFF
--- a/causal_testing/testing/causal_test_engine.py
+++ b/causal_testing/testing/causal_test_engine.py
@@ -162,7 +162,7 @@ class CausalTestEngine:
             )
         elif causal_test_case.estimate_type == "risk_ratio":
             logger.debug("calculating risk_ratio")
-            risk_ratio, confidence_intervals = estimator.estimate_risk_ratio()
+            risk_ratio, confidence_intervals = estimator.estimate_risk_ratio(estimator.params)
             causal_test_result = CausalTestResult(
                 estimator=estimator,
                 test_value=TestValue("risk_ratio", risk_ratio),
--- a/causal_testing/testing/causal_test_engine.py
+++ b/causal_testing/testing/causal_test_engine.py
@@ -162,10 +162,8 @@ class CausalTestEngine:
             )
         elif causal_test_case.estimate_type == "risk_ratio":
             logger.debug("calculating risk_ratio")
-            try:
-                risk_ratio, confidence_intervals = estimator.estimate_risk_ratio(causal_test_case.estimate_params)
-            except TypeError:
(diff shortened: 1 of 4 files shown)


causal_testing/testing/causal_test_engine.py
```diff
@@ -162,7 +162,10 @@ def _return_causal_test_results(self, estimator, causal_test_case):
             )
         elif causal_test_case.estimate_type == "risk_ratio":
             logger.debug("calculating risk_ratio")
-            risk_ratio, confidence_intervals = estimator.estimate_risk_ratio()
+            try:
```
jmafoster1, line 165: Is this to catch bad estimate parameters? If so, I'd suggest crashing out with an error rather than falling back on the defaults. If I'd supplied bad arguments, this is the behaviour I'd expect. Otherwise, the user could reasonably expect the returned result to be according to their supplied args when it's actually just from the defaults...

jmafoster1:
I personally think this is fairly sensible. It's how doWhy does it, and we need something which is flexible to different estimators, which may require different parameters.

causal_testing/testing/causal_test_engine.py
cwild-UoS: This allows for the fact that different estimators implement risk ratios differently and some implementations have no parameters. 

Good point though with it potentially hiding bad parameters and just selecting defaults. I'll look at adding an info logging line if default values are chosen, so there is some visibility?

cwild-UoS:
> I personally think this is fairly sensible. It's how doWhy does it, and we need something which is flexible to different estimators, which may require different parameters.

Good to know that is how another library does it. I must admit reading 'it's how doWhy does it' confused me for a minute 😅

causal_testing/testing/causal_test_engine.py
jmafoster1: If it's just intended for that, would it be better to `**` the arguments to the estimator constructor? Then if you end up `**`ing nothing, it should still work OK

cwild-UoS: Ahhh that looks a lot cleaner, cheers Micahel! 

I felt that I was missing something..

jmafoster1 (approved):

PR #213, opened 2023-07-14
Removed the big `if` statement in causal test engine

jmafoster1:
This requires a specific naming structure of causal effect measure calculation methods, e.g. if we want to estimate the "ate", our estimator must have a method `estimate_ate`, but we were basically following this anyway. This then enables us to get rid of the big `if` statement and use `getattr` instead.

### DIFF
--- a/causal_testing/testing/causal_test_engine.py
+++ b/causal_testing/testing/causal_test_engine.py
@@ -147,7 +147,7 @@ class CausalTestEngine:
         :return: a CausalTestResult object containing the confidence intervals
         """
         if not hasattr(estimator, f"estimate_{causal_test_case.estimate_type}"):
-            raise NotImplementedError(f"{estimator.__class__} has no {causal_test_case.estimate_type} method.")
+            raise AttributeError(f"{estimator.__class__} has no {causal_test_case.estimate_type} method.")
         estimate_effect = getattr(estimator, f"estimate_{causal_test_case.estimate_type}")
         effect, confidence_intervals = estimate_effect(**causal_test_case.estimate_params)
         causal_test_result = CausalTestResult(
--- a/causal_testing/testing/causal_test_engine.py
+++ b/causal_testing/testing/causal_test_engine.py
@@ -148,63 +148,17 @@ class CausalTestEngine:
         :param causal_test_case: The concrete test case to be executed
         :return: a CausalTestResult object containing the confidence intervals
         """
-        if causal_test_case.estimate_type == "cate":
-            logger.debug("calculating cate")
-            if not hasattr(estimator, "estimate_cates"):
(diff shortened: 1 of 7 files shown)


jmafoster1:
@cwild-UoS the macos tests seem to be failing because it "failed to properly upload". Please could you take a quick look if you get the chance?

cwild-UoS:
> @cwild-UoS the macos tests seem to be failing because it "failed to properly upload". Please could you take a quick look if you get the chance?

For now I've rerun the jobs without any changes as I've read a couple of places this fixes it 🤷.

If that fails I'll change the `secrets.CODECOV_TOKEN`, which isn't required for public repos but can fix this issue as well

jmafoster1:
Seems to be working now :)

causal_testing/testing/causal_test_engine.py
```diff
@@ -148,63 +146,17 @@ def _return_causal_test_results(self, estimator, causal_test_case):
...
-                confidence_intervals=confidence_intervals,
-            )
-            # causal_test_result = CausalTestResult(minimal_adjustment_set, ate, confidence_intervals)
-            # causal_test_result.apply_test_oracle_procedure(self.causal_test_case.expected_causal_effect)
-        else:
-            raise ValueError(
-                f"Invalid estimate type {causal_test_case.estimate_type}, expected 'ate', 'cate', or 'risk_ratio'"
-            )
+        if not hasattr(estimator, f"estimate_{causal_test_case.estimate_type}"):
+            raise NotImplementedError(f"{estimator.__class__} has no {causal_test_case.estimate_type} method.")
```
cwild-UoS, line 150: `NotImplementedError` I think is reserved for abstract methods. Here an `AttributeError` would be more appriopriate

causal_testing/testing/estimators.py
```diff
@@ -495,7 +495,7 @@ def add_modelling_assumptions(self):
             (iii) Instrument and outcome do not share causes
         """
 
-    def estimate_coefficient(self, df):
+    def estimate_coefficient_aux(self, df):
```
cwild-UoS, line 498: What does aux stand for here? It might not be obvious to users the difference between the `estimate_coefficient` and `estimate_coefficent_aux` methods

cwild-UoS (approved):
Looks great, neat way of implementing this imo 👍

causal_testing/testing/causal_test_engine.py
jmafoster1: Fixed

causal_testing/testing/estimators.py
jmafoster1: Auxiliary. I think this is semi-standard practice in functional programming, but is probably not "pythonic". I renamed the method to "estimate_iv_coefficient"

PR #219, opened 2023-07-26
Test engine refactor

cwild-UoS:


cwild-UoS:
Hi @jmafoster1 ,

I've marked it ready for review so you can start to look through the big changes here. There are one or two things to discuss in more detail I think? 

I'll fix the linting issues over the next couple of working days, but thought it would be good for you to start reviewing now :)

causal_testing/json_front/json_class.py
```diff
@@ -87,6 +87,8 @@ def setup(self, scenario: Scenario):
                 "attribute with a dataframe before calling .setup()"
             )
         self._populate_metas()
+        self.data_collector = ObservationalDataCollector(
```
jmafoster1, line 90: Hardcoding an `ObservationalDataCollector` seems dangerous. What if the user wants to use an `ExperimentalDataCollector`? We've only ever used one once, but that in the CTF paper to show that you can do it, so that means we need to keep support for it, unfortunately

causal_testing/json_front/json_class.py
```diff
@@ -246,7 +250,8 @@ def _populate_metas(self):
             meta.populate(self.data)
 
     def _execute_test_case(
-        self, causal_test_case: CausalTestCase, test: Iterable[Mapping], f_flag: bool
+            self, causal_test_case: CausalTestCase, test: Iterable[Mapping],
```
jmafoster1, line 253: Just noticed this, but why is `test` and `Iterable`?

causal_testing/testing/causal_test_case.py
```diff
@@ -73,6 +81,45 @@ def get_treatment_value(self):
...
 
+    def execute_test(self, estimator: type(Estimator), data_collector: ObservationalDataCollector) -> CausalTestResult:
+        """Execute a causal test case and return the causal test result.
+
+        :param estimator: A reference to an Estimator class.
+        :param causal_test_case: The CausalTestCase object to be tested
+        :return causal_test_result: A CausalTestResult for the executed causal test case.
+        """
+        if estimator.df is None:
+            estimator.df = data_collector.collect_data()
```
jmafoster1, line 92: Will this basically be dead code if we're calling `collect_data()` in the json frontend? If so, do we need to call it there? Could we just get away with calling it here instead?

causal_testing/testing/causal_test_case.py
```diff
@@ -73,6 +81,45 @@ def get_treatment_value(self):
...
+        treatment_variable = self.treatment_variable
+        treatments = treatment_variable.name
+        outcome_variable = self.outcome_variable
```
jmafoster1, line 95: Given that these are only used for logging, we could probably refactor them out and just use the `self` references.

causal_testing/testing/causal_test_suite.py
```diff
@@ -37,3 +43,45 @@ def add_test_object(
...
+    def execute_test_suite(self, data_collector: ObservationalDataCollector,
+                           causal_specification: CausalSpecification) -> list[CausalTestResult]:
+        """Execute a suite of causal tests and return the results in a list
+        :param test_suite: CasualTestSuite object
+        :return: A dictionary where each key is the name of the estimators specified and the values are lists of
+                causal_test_result objects
+        """
+        if data_collector.data.empty:
+            raise ValueError("No data has been loaded. Please call load_data prior to executing a causal test case.")
+        data_collector.collect_data()
```
jmafoster1, line 56: Another instance of `collect_data()`. Surely this one must be redundant?

causal_testing/testing/causal_test_suite.py
```diff
@@ -37,3 +43,45 @@ def add_test_object(
         """
         test_object = {"tests": causal_test_case_list, "estimators": estimators_classes, "estimate_type": estimate_type}
         self.data[base_test_case] = test_object
+
+    def execute_test_suite(self, data_collector: ObservationalDataCollector,
+                           causal_specification: CausalSpecification) -> list[CausalTestResult]:
+        """Execute a suite of causal tests and return the results in a list
```
jmafoster1, line 49: I'm a little confused with what this is actually doing? Surely this should just be a `for` loop now of `test.execute()`, or am I misunderstanding what this class was meant for. Also it looks to be returning a dict not a list, so either the docstring or the returned object should be updated.

causal_testing/json_front/json_class.py
cwild-UoS: Good spot, it should just be Mapping

causal_testing/testing/causal_test_case.py
cwild-UoS: I agree, calling it only in `causal_test_case` makes more sense, will fix :)

causal_testing/testing/causal_test_suite.py
cwild-UoS: Sorry for this making it in, I'm honestly not sure why it was using the `_return_causal_test_results` method rather than `execute_test`.

causal_testing/json_front/json_class.py
cwild-UoS: I'm not sure if there is a nice way of making the JSON frontend support both. 

The `ExperimentalDataCollector` has abstract methods that need to be defined somewhere, which goes against the JSON frontend design of being a command line utility. There is the additional issue of where would the parameters required to instantiate an `ExperimentalDataCollector` be specified.

cwild-UoS: If we do keep it as only `ObservationalDataCollector`s though, I think it is worth making it very clear that the frontend only supports that style

jmafoster1: Could we do the same thing as we did with estimators and that, and just have a dict to convert between strings and classes?

cwild-UoS: We could, but the user would still have to make a separate python script in order to implement the `run_system_with_input_configuration` method from the `ExperimentalDataCollector`. Then would need to manually invoke the Json frontend or copy and paste the argparser stuff across.

It just feels a bit messier, but doable if it is useful.

cwild-UoS: I'll make this an issue for a future PR for now, as this is already a big PR and it is a bit out of the scope. The previous JSON implementation also had this hard-coded so for now it isn't changing the functionality.

causal_testing/testing/causal_test_case.py
```diff
@@ -73,6 +77,41 @@ def get_treatment_value(self):
         """Return the treatment value of the treatment variable in this causal test case."""
         return self.treatment_value
 
+    def execute_test(self, estimator: type(Estimator), data_collector: ObservationalDataCollector) -> CausalTestResult:
```
jmafoster1, line 80: Should `ObservationalDataCollector` just be `DataCollector`? Otherwise there's no way to execute a test with an `ExperimentalDataCollector` with or without the JSON front end.

PR #223, opened 2023-08-02
Add wheel dependency to GH action

cwild-UoS:
New versions aren't being pushed to PyPI due to wheel dependency is not being found.

Not sure why wheel is suddenly required explicitly.


cwild-UoS:
Cancelled CI tests as no python code changed

jmafoster1 (approved):

PR #225, opened 2023-08-07
add estimator params for linear regression estimate methods

cwild-UoS:
For issue #203


causal_testing/testing/estimators.py
```diff
@@ -393,26 +392,33 @@ def estimate_control_treatment(self, adjustment_config: dict = None) -> tuple[pd
 
         return y.iloc[1], y.iloc[0]
 
-    def estimate_risk_ratio(self) -> tuple[float, list[float, float]]:
+    def estimate_risk_ratio(self, estimator_params: dict = None) -> tuple[float, list[float, float]]:
```
jmafoster1, line 395: What about the other estimators, for example in `LogisticRegressionEstimator` and `CausalForrestEstimator`? Should they also be migrated to this `estimator_params` paradigm? Would a more elegant solution be to call `estimate_ate` with `**estimator_params` rather than passing the dict in directly? Then we could keep all the keyword arguments rather than having to specify the params in a dict, which seems a bit horrible for those calling the method directly

cwild-UoS: Ofc.. That is a lot cleaner approach. I've reverted to that design now.

One concern with this approach of unpacking `estimator_params` like `**estimator_params`, is that there is currently no validation on spelling etc for those params. Because if the parameter is not found it will simply select the default value. 

We could collect the `**kwargs` in each method like `estimate_ate` and check that it is empty? This way if there were any parameters in the `estimate_params` that weren't used we could raise an exception.

jmafoster1: That sounds like a sensible idea, although would it not kick off with an "unexpected keyword argument" type error anyway if it didn't recognise it?

jmafoster1: Great! I'll leave it up to @rsomers1998 to make the approving review, so he can check this PR will actually allow him to do what he wants

rsomers1998 (approved):
lgtm. Cheers for this

PR #232, opened 2023-08-15
Fixed serialisation bug

jmafoster1:
Fixed a problem with the `to_json` method of causal test result where pandas series weren't json serialisable. Now converting them to a dict.

### DIFF
--- a/causal_testing/testing/causal_test_result.py
+++ b/causal_testing/testing/causal_test_result.py
@@ -85,9 +85,11 @@ class CausalTestResult:
             "outcome": self.estimator.outcome,
             "adjustment_set": list(self.adjustment_set) if json else self.adjustment_set,
             "effect_measure": self.test_value.type,
-            "effect_estimate": self.test_value.value,
-            "ci_low": self.ci_low(),
-            "ci_high": self.ci_high(),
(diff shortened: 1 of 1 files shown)


cwild-UoS:
According to the typing, `test_value.value` and the confidence intervals should be floats, which don't have a `to_dict()` method. 

What types are they in the scenario you are running?

jmafoster1:
They are mostly floats, but are sometimes pandas `Series` objects. This happens when you have a treatment (or variables in the adjustment set) which are categorical. Then, instead of a float, you get a float for each value, which comes as a `Series`.

cwild-UoS:
> Fixed a problem with the `to_json` method of causal test result where pandas series weren't json serialisable. Now converting them to a dict.

Ahh okay that makes sense. Sorry for more questions, but in your original comment you mention a `to_json` method. Where is this method? I wasn't aware we were serialising the `causal_test_result` objects

jmafoster1:
Ah sorry, I think it's actually called `to_dict` in the `CausalTestResult` class (see Files Changed). I use it to serialise the causal test results to JSON so I can create a better log than the default one that I can just parse straight into my results processing program.

cwild-UoS (approved):

PR #233, opened 2023-08-15
Remove Redundant Adjustment Set Calc when Formula

### LINKED ISSUE #206: Redundant adjustment set calculation
Recently, I've been doing a lot more work with custom formulae, adding in squared terms, etc. This works well, but the CTF still calculates the adjustment set in the traditional way, which is then overruled by my custom formula. Firstly, this is a bit dangerous since there is no validation that the formula represents a correct adjustment set, so it would be good to validate this. Secondly, the adjustment set is quite intensive and slow to calculate (even with the new algorithm), so it would be good to not have to do this where a formula has been provided. This would speed up test execution by quite a lot, I think.

cwild-UoS:
This PR removes the unnecessary adjustment set calculation when a formula is provided with a regression estimator. Discussed in #206. Doesn't close #206 as also need to implement checks on the formula adjustment set


cwild-UoS:
For now I've got it to raise a `TypeError` if a formula is specified for an estimator that isn't a `LinearRegressionEstimator`. 

Is it likely that any other estimators in the future will want to use formulas?

causal_testing/json_front/json_class.py
```diff
@@ -307,20 +307,26 @@ def _setup_test(self, causal_test_case: CausalTestCase, test: Mapping) -> Estima
...
-            "treatment_value": causal_test_case.treatment_value,
-            "control_value": causal_test_case.control_value,
-            "adjustment_set": minimal_adjustment_set,
-            "outcome": causal_test_case.outcome_variable.name,
-            "effect_modifiers": causal_test_case.effect_modifier_configuration,
-            "alpha": test["alpha"] if "alpha" in test else 0.05,
-        }
+        estimator_kwargs = {}
         if "formula" in test:
+            if test["estimator"] != LinearRegressionEstimator:
```
jmafoster1, line 312: `LogisticRegressionEstimator` does too!

causal_testing/testing/estimators.py
```diff
@@ -306,9 +306,6 @@ def __init__(
             terms = [treatment] + sorted(list(adjustment_set)) + sorted(list(effect_modifiers))
             self.formula = f"{outcome} ~ {'+'.join(terms)}"
 
-        for term in self.effect_modifiers:
```
jmafoster1, line 309: This still needs to be done if no formula is provided, or is this take account of elsewhere?

cwild-UoS: I wanted to ask about these lines! I wasn't sure where/if the adjustment set was used after this. 

None of the unittests seem to cover this being used either so I'll raise that as an issue and re-add the lines :).

jmafoster1:
For the validation of formulae, you can use the `constructive_backdoor_criterion` of the `CausalDAG` class. This will return `True` if your adjustment set is valid. Then it is just a matter of extracting the variables from the formula, which I would hope Patsy makes at least possible...

causal_testing/testing/estimators.py
jmafoster1: Yes, effect modifiers need to be added to the adjustment set. Causally, they are semantically different to normal adjustment variables but, at least for LR, they are treated the same way when you do the maths.

causal_testing/json_front/json_class.py
```diff
@@ -307,20 +307,29 @@ def _setup_test(self, causal_test_case: CausalTestCase, test: Mapping) -> Estima
...
-        }
+        estimator_kwargs = {}
         if "formula" in test:
+            if test["estimator"] != (LinearRegressionEstimator or LogisticRegressionEstimator):
+                raise TypeError(
+                    "Currently only LinearRegressionEstimator and LogisticRegressionEstimator supports the use of "
+                    "formulas"
+                )
             estimator_kwargs["formula"] = test["formula"]
+            estimator_kwargs["adjustment_set"] = {}
```
jmafoster1, line 318: Please can this be `None` instead of empty? A minor thing, but it would help to distinguish the causal semantics of "we don't need an adjustment set because there is no confounding" and "we don't provide an adjustment set because we provide a formula instead"

jmafoster1 (approved):

PR #235, opened 2023-08-21
Formula validation

cwild-UoS:


causal_testing/json_front/json_class.py
```diff
@@ -270,6 +270,9 @@ def _execute_test_case(
         failed = False
 
         estimation_model = self._setup_test(causal_test_case=causal_test_case, test=test)
+        if "formula" in test:
+            if not estimation_model.validate_formula(self.causal_specification.causal_dag):
+                raise ValueError("Formula covariates do not satisfy the constructive back-door criterion.")
```
jmafoster1, line 275: Given that this is what this PR is meant to do, I think it'd be good to have this line covered by a test to make sure that invalid adjustment sets really are picked up. If you have a DAG "X -> Y; Z -> X; Z -> Y" and a causal test with formula "Y ~ X", that should set it off.

causal_testing/testing/estimators.py
```diff
@@ -100,16 +101,97 @@ def __init__(
...
+        outcome = desc.lhs_termlist[0].factors[0].code
+        rhs_terms = set()
+        for term in desc.rhs_termlist:
+            if term.factors:
+                rhs_terms.add(term.factors[0].code)
+        if self.treatment not in rhs_terms:
+            raise ValueError(f"Treatment variable '{self.treatment}' not found in formula")
+        rhs_terms.remove(self.treatment)
+        covariates = rhs_terms
+        if covariates is None:
```
jmafoster1, line 147: Will this ever actually come about? Surely there should always be at least one RHS term, even if it's just the treatment?

cwild-UoS: The code link isn't too clear on if you are talking about the ValueError or the `if covariates is None`,

The `if covariates is None:` addresses the case where there is only the treatment variable on the RHS and therefore there are no covariates, but the `constructive_backdoor_criterion` expects a list instead of `None`. 

The `ValueError` covers the case where the formula does not include the treatment hopefully.

Let me know if that clears it up?

jmafoster1: Ah sorry. I was going for `if covariates is None`, and that makes sense. Again, a test to cover this would probably be good. Causal DAG `X -> Y` with formula `Y ~ X` should do it

f-allian (approved):
Approving changes

PR #240, opened 2023-08-31
Fixed linalg error

f-allian:
@jmafoster1's temporary fix for the lingalg errors in v5.2.1.


jmafoster1:
Thanks for this @f-allian. If I merge this into main and create a new release, will that be OK? Should I call it 5.2.2?

jmafoster1:
I've also just noticed that it says the branch is out of date with `main`, presumably because I cloned v5.2.1. Thus, I don't want to update to `main` because that will, essentially, be cloning from 5.3.0, which I don't want to do. Is there a resolution to this?

f-allian:
> I've also just noticed that it says the branch is out of date with `main`, presumably because I cloned v5.2.1. Thus, I don't want to update to `main` because that will, essentially, be cloning from 5.3.0, which I don't want to do. Is there a resolution to this?

@jmafoster1 I've just noticed this too. Are you sure you cloned v5.2.1 and not v5.3.0? From what I can see in the change logs, the 17 changes that want to be merged comes from v5.3.0 (e.g. see [this commit](https://github.com/CITCOM-project/CausalTestingFramework/commit/a2abf7b54970898b2633a3f2301d312ff0dc9d57) and this [change log](https://github.com/CITCOM-project/CausalTestingFramework/compare/v5.2.1...v5.3.0)). I think, in this case, the safest thing for you to do is to make a new clone of the main in this current version (v5.2.1), and open another PR. 

I'm not sure if there's a better/safer alterative!

Once you do this, you can create a new release (v5.2.2 is fine) using the [notes](https://causal-testing-framework.readthedocs.io/en/latest/dev/version_release.html) Chris made here :)

jmafoster1 (approved):

PR #241, opened 2023-09-18
Fixed "np.int64" is not JSON encodable from test outcomes

jmafoster1:
Small fix to the Positive and Negative test outcomes such that they return native `bool` datatypes rather than `np.bool`, resulting in `np` datatypes being carried forward into other parts of the CTF and causing JSON encoding issues further down the line.


f-allian (approved):
Approving changes

PR #244, opened 2023-11-06
Update: edited example_vaccine.py

f-allian:
Replaced the `ExperimentalDataCollector` class in the covasim example with the  `ObservationalDataCollector` class to improve execution time of this example.


examples/covasim_/vaccinating_elderly/README.md
```diff
@@ -13,6 +13,9 @@ four test cases: one focusing on each of the four previously mentioned outputs.
 
 Further details are provided in Section 5.3 (Prioritising the elderly for vaccination) of the paper.
 
+Note: this version of the example utilises the observational data collector, which was previously simulated using the 
+experimental data collector, in order to save computation time.
```
cwild-UoS, line 17: The experimental data collector is being removed from the framework as it is more logical to separate the software execution and testing. It may be worth rephrasing this to avoid future confusion

examples/covasim_/vaccinating_elderly/example_vaccine.py
```diff
@@ -1,38 +1,34 @@
...
-from causal_testing.data_collection.data_collector import ExperimentalDataCollector
+from causal_testing.data_collection.data_collector import ObservationalDataCollector
 from causal_testing.testing.causal_test_case import CausalTestCase
 from causal_testing.testing.causal_test_outcome import Positive, Negative, NoEffect
 from causal_testing.testing.estimators import LinearRegressionEstimator
 from causal_testing.testing.base_test_case import BaseTestCase
 
-import os
-import logging
+warnings.filterwarnings("ignore", message=".*The 'nopython' keyword.*")  # Supress the numba warning in covasim
```
cwild-UoS, line 16: Is this needed now covasim is removed?

examples/covasim_/vaccinating_elderly/example_vaccine.py
```diff
@@ -1,38 +1,34 @@
...
+    data collector that was previously simulated.
 
-    :param runs_per_test_per_config: Number of times to run each input configuration (control and treatment) per test.
-                                     Hence, the total number of runs per test will be twice this value.
     :param verbose: Whether to print verbose details (causal test results).
     :return results_dict: A dictionary containing ATE, 95% CIs, and Test Pass/Fail
     """
 
-    # 1. Read in the Causal DAG
+    # 1. Create causal dag
```
cwild-UoS, line 31: To me the previous comment was more clear, is there a reason for this change?

cwild-UoS (changes requested):
Sorry doing this on my phone, will do a more thorough look later today :).

examples/covasim_/vaccinating_elderly/example_vaccine.py
f-allian: Thanks for your suggestions @cwild-UoS - I'll review them this morning. 

For some reason, the Numba warnings are still outputed even though covasim isn't directly called in this script. I wonder whether they're called indirectly through another module? But it does seem to be needed to supress the warnings.

examples/covasim_/vaccinating_elderly/example_vaccine.py
f-allian: There isn't at all - no idea why I changed the original comment!

examples/covasim_/vaccinating_elderly/example_vaccine.py
cwild-UoS: Have you run the script in an environment that doesn't have Covasim installed? Just curious if that would change anything

f-allian: @cwild-UoS I traced it back to EconML - the two modules that use Numba are Shap and Sparse, which are both implicitly installed. Do we need EconML? I can't find any usage of it in the framework

cwild-UoS: I can't recall what EconML was used for off the top of my head :(. Might be worth going back to an earlier version and seeing where it was used and when it stopped being used? 

If there definitely are no uses then I can't see why it can't also be removed.

f-allian: @cwild-UoS It was only used in the `CausalForestEstimator`. This also ties in with the Python 3.11 branch you were working on, as you removed EconML from the .toml file in your commits

cwild-UoS (approved):
Looks good to me 👍

PR #247, opened 2023-11-14
Add workflow to release to Figshare on release

cwild-UoS:
The secret is already made.

The record is still private in my workspace until this workflow is tested so the first DOI is attached to the first push


f-allian (approved):
Nice one Chris, thanks for your work on this - looks good.

cwild-UoS:
Cheers! I'll close the issue once the workflow is tested :)!

PR #248, opened 2023-11-15
Update figshare.yaml

cwild-UoS:
Figshare article number wasn't an integer


f-allian (approved):
Ah sorry I missed this too. I wasn't sure if `<<>>` was the syntax for the variable

PR #249, opened 2023-11-16
Update: docs revamp

### LINKED ISSUE #242: Update and Improve the Documentation
We should really update our documentation (ideally on a regular cadence, e.g. automating when every release is made via webhooks?) to reflect recent changes. For instance, the [usage section](https://causal-testing-framework.readthedocs.io/en/latest/usage.html) still references the causal_test_engine, which has been refactored for a while now.

Also, a couple of (semantic) points worth discussing regarding separate feature requests:

1. The documentation IMO is generally quite text-heavy - can/should we introduce more diagrams, flow-charts, etc. to make the framework more accessible for non-expert programmers, and making it more aesthetic? 
2. Similarly, do we want to choose a theme for the documentation? Something specific for the CITCOM branding (if one exists)?

If you have any more ideas/suggestions, please do add them in this thread, and I'm happy to own this issue.

f-allian:
Closes #242


cwild-UoS:
Small thing but I also get this glossary warning

```
C:\Users\cs1xcw\PycharmProjects\CausalTestingFramework\docs\source\index.rst:9: WARNING: term not in glossary: 'directed acyclic graphs'
```

Not sure if this is a problem or not though

cwild-UoS (changes requested):
Forgot to add my comments as a review so I'll add this. Like we discussed though no rush on this at all

f-allian:
> Forgot to add my comments as a review so I'll add this. Like we discussed though no rush on this at all. It just means when you push the changes you can re-request a review and I'll get pinged :)

Thanks for the feedback @cwild-UoS. That should be all fixed now - you'll likely still receive the usual autoapi warnings, though.

cwild-UoS (approved):
Looks great to me! Cheers for doing this Farhad

Issue #92, opened 2022-09-27
DAFNI / Docker


bobturneruk:
It would be good to see if we can get causal testing framework running on [DAFNI](https://www.dafni.ac.uk/). This would entail containerising code, but work needs to be done to determine what the best way to do this is. Do we offer a standard containerised framework, and / or instructions to containerise testing for particular model outputs? How do we pass variables to and from containers.

PR #251, opened 2023-12-12
Dockerised framework for DAFNI

f-allian:
This PR contains an initial commit for the dockerisation of the causal testing framework, so that it can be hosted on [DAFNI](https://www.dafni.ac.uk/) - closes issue #92 after review.

## Steps to reproduce tests

This PR uses the [vaccinating elderly](https://github.com/CITCOMproject/CausalTestingFramework/tree/main/examples/covasim_/vaccinating_elderly) example as a test-case. The entry point to the framework is wrapped in a script called `main_dafni.py`, which takes in 4 mandatory input arguments, including `data_path`, `dag_path`, `tests_path`, `variables_path`.

### 1. Without Docker

Simply run:

`python main_dafni.py --variables_path $VARIABLES --dag_path $DAG_PATH --data_path $DATA_PATH --tests_path $CAUSAL_TESTS`,

and point to the path containing your configuration files (for this example, everything is defined under `./dafni/inputs`). The resultant causal tests will be saved in `.json` format in the folder `./dafni/outputs`. (Note; the folder structure here is important for DAFNI)

### 2. With Docker

1. Create a `.env` file in the `./dafni` directory containing the environment variables, which is then passed into the dockerfile for the build.
2. Then let docker-compose do the building of the image and creating the container by simply running `docker-compose up`.

Overall, the total execution time (building of the image and script execution) takes ~1 minute on my computer (this may take slightly longer on a new setup as there won't be any cached data).

## Overall Progress

- [x] #92
- [x] Create a wrapper of the causal testing framework to be used as an entry-point for Docker
- [x] Containerise the causal testing framework using Docker
- [x] Update the  `model_definition.yaml` to contain the appropriate dataslot IDs (if any) needed for the execution of the framework
- [x] Conduct more tests using different examples (we should have plenty to test)
- [x] Upload to DANFI.


dafni/inputs/variables.json
jmafoster1, line 1: Perhaps I'm just not software engineering-y enough, but why do we need separate files for inputs and tests, each with a single key in for which the value is a list? Could we not either combine `tests.json` and `variables.json` into a single file, or have them each just contain a list of the values, or is that in some way bad practice?

jmafoster1: A potential alternative to this would be to include this metadata in the dag file as attributes for each variable. That could make things simpler from a user's point of view as they'd then only have to make 1 file (the DAG, the tests can be built automatically from the DAG).

dafni/main_dafni.py
```diff
@@ -0,0 +1,210 @@
...
+
+        estimators = {"LinearRegressionEstimator": LinearRegressionEstimator}
+
+        # Step 3: Define the expected variables
+
+        expected_outcome_effects = {
+            "Positive": Positive(),
+            "Negative": Negative(),
+            "NoEffect": NoEffect(),
+            "SomeEffect": SomeEffect()}
```
jmafoster1, line 166: Is there any way we can get round this hardcoding? This makes it pretty much impossible for a user to implement customisations, and is also the only blocker to having a single default `main.py` file (I think), now we're putting variables in a file.

jmafoster1:
@f-allian please can you post your `.env` file as an example? I've never made one before, so guessed at the following, but `docker-compose up` fails with error message ` main_dafni.py: error: argument --variables_path: expected one argument`
My `.env` file:
```
VARIABLES=dafni/variables.json
DAG_PATH=dafni/dag.dot
DATA_PATH=dafni/simulated_data.csv
CAUSAL_TESTS=dafni/causal_tests.json
```

cwild-UoS:
Hi Michael,

I believe your `VARIABLES` line should instead be `VARIABLES_PATH`. i.e.

```
VARIABLES_PATH=inputs/variables.json
DAG_PATH=inputs/dag.dot
DATA_PATH=inputs/simulated_data.csv
CAUSAL_TESTS=inputs/causal_tests.json
```

jmafoster1:
Thanks Chris, that works. Would it be sensible to commit the `.env` file as part of the example, or is there a reason why this isn't a good idea?

f-allian:
> Thanks Chris, that works. Would it be sensible to commit the `.env` file as part of the example, or is there a reason why this isn't a good idea?

@jmafoster1 There isn't a specific reason why I didn't include it in my commit in this case. Generally speaking, environment files typically contain passwords, API keys etc., which is why they're left out of the version controlling stage. I don't think DAFNI requires that to be specified in our case, so I can provide a template if you think that's useful. 

Also, if you think some of my variable naming conventions aren't helpful/easily identifiable, please let me know!

jmafoster1:
I think it would be helpful to show an example/template. I'm happy with variable names, but we should update the "without docker" run command to include the `_PATH` suffix to be consistent.

dafni/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+COPY --chown=nobody ./dafni/main_dafni.py /usr/src/app/
+
+# Change the working directory
+WORKDIR /usr/src/app/
+
+# Install core dependencies using PyPi
+RUN pip install causal-testing-framework --no-cache-dir
+
+# Use the necessaary environment variables for the script's inputs
+ENV VARIABLES=./inputs/variables.json \
```
rsomers1998, line 26: I understand that these are overwritten by the .env but `VARIABLES` here should be `VARIABLES_PATH`. Also, if we're assuming a .env to be supplied, are setting these within the dockerfile necessary?

dafni/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+# Copy the source code and test files from build into the container
+COPY --chown=nobody ./causal_testing /usr/src/app/
+COPY --chown=nobody ./dafni/inputs /usr/src/app/inputs/
+COPY --chown=nobody ./dafni/main_dafni.py /usr/src/app/
+
+# Change the working directory
+WORKDIR /usr/src/app/
+
+# Install core dependencies using PyPi
+RUN pip install causal-testing-framework --no-cache-dir
```
cwild-UoS, line 23: Out of curiosity why is the `--no-cache-dir` flag used here? My guess would be that that there is no `causal-testing-framework` wheel in cache as it's a fresh container.

dafni/main_dafni.py
```diff
@@ -0,0 +1,210 @@
+import warnings
+warnings.filterwarnings("ignore", message=".*The 'nopython' keyword.*")
```
cwild-UoS, line 2: What warning does this filter out? I tried running the script without this line and didn't see any warnings.

dafni/main_dafni.py
```diff
@@ -0,0 +1,210 @@
...
+def read_variables(variables_path: Path) -> dict:
+    """
+    Function to read the variables.json file specified by the user
+    :param variables_path: A Path object of the user-specified file path
+    :returns:
+            - dict - A valid dictionary consisting of the causal tests
+    """
+    if not variables_path.exists() or variables_path.is_dir():
+
+        raise ValidationError(f"Cannot find a valid settings file at {variables_path.absolute()}.")
```
cwild-UoS, line 96: I'd argue this could be the built in `FileNotFoundError ` rather than a custom exception.

dafni/main_dafni.py
```diff
@@ -0,0 +1,210 @@
...
+
+        inputs = [Input(variable["name"], eval(variable["datatype"])) for variable in variables if
+                  variable["typestring"] == "Input"]
+
+        outputs = [Output(variable["name"], eval(variable["datatype"])) for variable in variables if
+                   variable["typestring"] == "Output"]
+
+        constraints = set()
+
+        for variable, _inputs in zip(variables, inputs):
```
cwild-UoS, line 126: Why is the element of `inputs` called `_inputs` rather than `input`? Typically the underscore convention is used for private variables

dafni/main_dafni.py
```diff
@@ -0,0 +1,210 @@
...
+
+        print(f"Cannot validate the specified input configurations: {ve}")
+
+    else:
+
+        print(f"Execution successful. Output file saved at {Path(args.output_path).parent.resolve()}")
+
+
+if __name__ == "__main__":
+    main()
```
cwild-UoS, line 210: For some old fashioned reason (joe heffer explained it once and I forgot), python files should end with a blank line haha

cwild-UoS (commented):
Looks really really good! Found some small clean ups

dafni/Dockerfile
f-allian: @cwild-UoS The ``--no-cache-dir`` option disables the downloading and storing of cached packages, which reduces the overall Docker image size. I haven't calculated what that difference is, but it's not too important for our purposes

dafni/main_dafni.py
f-allian: @cwild-UoS ``input`` is already a pre-defined method in Python, so it's not a good idea to overwrite it! I don't think the variable name matters too much in this case, but I can change it if you think it's needed

cwild-UoS: I think changing it from having an underscore would be good just to give off the right intent. Also isn't it already non pre-defined being `inputs` rather than `input`?

dafni/Dockerfile
cwild-UoS: Makes lots of sense! Good idea

dafni/Dockerfile
f-allian: @rsomers1998 You were indeed correct! Thanks for flagging this.

PR #257, opened 2024-01-30
We now support placing conditions on the data again.

jmafoster1:
Somehow the ability to place conditions on the data got lost at some point, so I have reintroduced it.


f-allian (approved):
@jmafoster1 looks good to me - there's a pylint error in estimators.py in the line `from patsy import dmatrix` for some reason, which is out of our hands. You can disable pylint picking this up by including `# pylint: disable = no-name-in-module` next to this line. I'll approve it anyway but might be worth changing this before merging.

PR #264, opened 2024-02-13
Estimator return types

cwild-UoS:
Fetch factors from interaction terms using Patsy ModelDesc class


cwild-UoS:
Added logic to handle both `pd.Series` and `float` values for confidence intervals. This is due to the fact that not all estimators return `pd.Series` and enforcing it on all of them feels awkward

cwild-UoS:
Different estimators and estimate methods (`estimate_ate`, `estimate_coefficient`, `estimate_risk_ratio`, etc) returning both floats of different kinds (`np.float64` and inbuilt `float`), and `pd.Series`, has lead to extra conditions being added in `causal_test_result.py` and `causal_test_outcome.py`

causal_testing/testing/causal_test_outcome.py
```diff
@@ -57,7 +58,7 @@ def apply(self, res: CausalTestResult) -> bool:
             ci_low = res.ci_low() if isinstance(res.ci_low(), Iterable) else [res.ci_low()]
             ci_high = res.ci_high() if isinstance(res.ci_high(), Iterable) else [res.ci_high()]
             value = res.test_value.value if isinstance(res.ci_high(), Iterable) else [res.test_value.value]
-
+            value = value[0] if isinstance(value[0], pd.Series) else value
```
jmafoster1, line 61: Please can you explain why this was necessary and nothing for the confidence intervals? I think it should have worked as it was before. If not, I'm surprised the confidence intervals didn't break anything

tests/testing_tests/test_estimators.py
```diff
@@ -231,7 +231,7 @@ def test_program_11_2(self):
         self.assertEqual(round(model.params["Intercept"] + 90 * model.params["treatments"], 1), 216.9)
 
         # Increasing treatments from 90 to 100 should be the same as 10 times the unit ATE
-        self.assertEqual(round(model.params["treatments"], 1), round(ate, 1))
+        self.assertEqual(round(model.params["treatments"], 1), round(ate[0], 1))
```
jmafoster1, line 234: If we're returning pandas series now, would it be better to do an assert all rather than just grabbing the first one? Obviously it'll be the same result since the series are all one element long, but I think it would capture the vibe of what we're going for better

causal_testing/testing/causal_test_outcome.py
jmafoster1: After looking at the test files, I also think this should probably work in some sort of loop too, rather than just grabbing the first value

cwild-UoS:
I may have to put this PR on hold for a week or 2. I've been at it on and off today / yesterday but really need to step back for my other projects. 

The outstanding work is fixing a couple of the examples and also Richards surrogate module. It turns out the fixing all the estimator methods to return only `pd.Series` was far more involved than I would have thought!

If we want this pushed sooner than I can get around to it we can have a chit on the google space 😄

jmafoster1:
Sure, no worries. I'm super busy at the moment anyway!

causal_testing/surrogate/surrogate_search_algorithms.py
```diff
@@ -46,7 +46,7 @@ def fitness_function(ga, solution, idx): # pylint: disable=unused-argument
 
                 ate = surrogate.estimate_ate_calculated(adjustment_dict)
 
-                return contradiction_function(ate)
+                return contradiction_function(ate[0])
```
jmafoster1, line 49: I'm a little worried about the scalability of this as it will only work for float values. If you have categorical data types, this will reduce to a binary treatment between the first and second values in alphabetical order. If you really want to only support float values for this, it would be good to have an explicit check for this to make sure that the user is getting what they expect to get.

causal_testing/testing/causal_test_outcome.py
```diff
@@ -99,10 +95,9 @@ def apply(self, res: CausalTestResult) -> bool:
         if res.ci_valid() and not super().apply(res):
             return False
         if res.test_value.type in {"ate", "coefficient"}:
-            return bool(res.test_value.value > 0)
+            return bool(res.test_value.value[0] > 0)
```
jmafoster1, line 98: Again, this breaks scalability to categorical data types, although this one is a bit more complicated. For a "positive" effect, you need an order. For now, we could just support float datatypes. To make sure we're not losing any information, we could raise a `TypeError` if `len(res.test_value.value) > 1`?

causal_testing/testing/causal_test_outcome.py
```diff
@@ -99,10 +95,9 @@ def apply(self, res: CausalTestResult) -> bool:
...
         if res.test_value.type == "risk_ratio":
-            return bool(res.test_value.value > 1)
-        # Dead code but necessary for pylint
+            return bool(res.test_value.value[0] > 1)
```
jmafoster1, line 100: See above

causal_testing/testing/causal_test_outcome.py
```diff
@@ -113,8 +108,8 @@ def apply(self, res: CausalTestResult) -> bool:
         if res.ci_valid() and not super().apply(res):
             return False
         if res.test_value.type in {"ate", "coefficient"}:
-            return bool(res.test_value.value < 0)
+            return bool(res.test_value.value[0] < 0)
```
jmafoster1, line 111: And here

causal_testing/testing/causal_test_outcome.py
```diff
@@ -113,8 +108,8 @@ def apply(self, res: CausalTestResult) -> bool:
...
         if res.test_value.type == "risk_ratio":
-            return bool(res.test_value.value < 1)
+            return bool(res.test_value.value[0] < 1)
```
jmafoster1, line 113: And here

causal_testing/testing/causal_test_result.py
```diff
@@ -27,7 +27,7 @@ def __init__(
         self,
         estimator: Estimator,
         test_value: TestValue,
-        confidence_intervals: [float, float] = None,
+        confidence_intervals: [pd.Series, pd.Series] = None,
```
jmafoster1, line 30: Would it make sense to switch to two instance variables, `ci_low` and `ci_high`?

tests/testing_tests/test_causal_test_case.py
```diff
@@ -118,7 +118,7 @@ def test_execute_test_observational_causal_forest_estimator(self):
             self.df,
         )
         causal_test_result = self.causal_test_case.execute_test(estimation_model, self.data_collector)
-        self.assertAlmostEqual(causal_test_result.test_value.value, 4, delta=1)
+        self.assertAlmostEqual(causal_test_result.test_value.value[0], 4, delta=1)
```
jmafoster1, line 121: Just a readability thing, but would it be worth doing something like `self.assertAlmostEqual(causal_test_result.test_value.value, pd.Series(4), delta=1)` instead to show that we expect the result to be a single element series?

cwild-UoS: I really like that approach! I'll go through this afternoon and update any cases to that 👍

causal_testing/testing/causal_test_outcome.py
cwild-UoS: For a categorical it isn't enough to just check that every variable is positive (greater than 0)? 

The suggestion of raising an error seems very logical to me to avoid unexpected behaviour though for now

jmafoster1: Does this not just check the first one though? I'd have expected to see an `all` in there somewhere...

causal_testing/testing/causal_test_result.py
cwild-UoS: I think this will make accessing the confidence intervals clearer. I've made it an issue to hopefully follow up on in another PR:
https://github.com/CITCOM-project/CausalTestingFramework/issues/266

causal_testing/testing/causal_test_outcome.py
cwild-UoS: Sorry should have been clearer! It does currently only check the first one, but if I were to change it to check that all values are positive would that be enough?

I'm not sure what you meant by 'you need an order'.

jmafoster1: In this case, I don't think so. A positive relationship here means "as we increase treatment X, we expect to see a corresponding increase in outcome Y". For this to apply, we need the value of X to be orderable such that "increasing X" is meaningful. If we just assert that all values are positive, we are just asserting that whatever value statsmodels uses as its reference value is the one for which we observe the minimal outcome value (and thus that all other treatment values would lead to a higher outcome). For now, I think it's better to just assert that there's only one value and then just index it. Perhaps we could add this explanation as part of the documentation for posterity.

tests/testing_tests/test_causal_test_case.py
cwild-UoS: I hadn't noticed that these tests use `self.assertAlmostEqual` which can't compare `pd.Series` objects. 

I think any attempt to add the semantics of the test result being a `pd.Series`, would add too much complexity to the test.

jmafoster1: We could just round the values and then use `self.assertEqual`?

cwild-UoS: I don't think you can make any assertions directly on pandas objects using only pytest. I think we would have to employ something such as [this](https://pandas.pydata.org/docs/reference/api/pandas.testing.assert_series_equal.html).

jmafoster1: That's quite nice, though, I think. The last one I had to do, I just did `assertTrue(all(x1==x2))`, which doesn't provide a nice error message

jmafoster1 (approved):

PR #265, opened 2024-02-22
Test categorical confidence intervals

jmafoster1:
Added a test for the `LinearRegressionEstimator` which tests confidence intervals as a pandas series for a categorical data type. Ideally, the `estimate_*` methods should all return pandas series for confidence intervals.


cwild-UoS (approved):
Wonderful, thank you for taking the time to make this test Michael!

cwild-UoS:
Merging now so I can make use of in #264

PR #267, opened 2024-03-06
Fix bug in categorical gene types

rsomers1998:
Pygad High label is exclusive. categorical variables should account for this

### DIFF
--- a/causal_testing/surrogate/surrogate_search_algorithms.py
+++ b/causal_testing/surrogate/surrogate_search_algorithms.py
@@ -107,6 +107,8 @@ class GeneticSearchAlgorithm(SearchAlgorithm):
                         var_space[rel_split[0]]["high"] = int(rel_split[2]) + 1
                     else:
                         var_space[rel_split[0]]["high"] = datatype(rel_split[2])
+            else:
+                raise ValueError(f"Value {rel_split[0]} not found in var_space")
 
         gene_space = []
         gene_space.append(var_space[surrogate_model.treatment])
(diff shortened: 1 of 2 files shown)


causal_testing/surrogate/surrogate_search_algorithms.py
```diff
@@ -99,10 +99,14 @@ def create_gene_types(
...
+                datatype = specification.scenario.variables.get(rel_split[0]).datatype
                 if rel_split[1] == ">=":
-                    var_space[rel_split[0]]["low"] = int(rel_split[2])
+                    var_space[rel_split[0]]["low"] = datatype(rel_split[2])
                 elif rel_split[1] == "<=":
-                    var_space[rel_split[0]]["high"] = int(rel_split[2])
+                    if datatype == int:
+                        var_space[rel_split[0]]["high"] = int(rel_split[2]) + 1
+                    else:
+                        var_space[rel_split[0]]["high"] = datatype(rel_split[2])
```
f-allian, line 109: @rsomers1998 Looks good to me overall. But from an error handling point of view, it'd be helpful to close all of your if-statements using relevant exceptions, e.g.:

```
if rel_split[0] in var_space:
 ....
else:
      Raise ValueError("Value not found in var_space")```

f-allian (changes requested):
Just a minor tweak:

causal_testing/surrogate/surrogate_search_algorithms.py
rsomers1998: A variable may not show up as the treatment variable or in the adjustment set but may still have constraints. This variable may be relevant to another causal relationship under test but not the one currently being tested. In this case, those constraints don't need to be fed into the search algorithm so can just be ignored. 

Throwing an error would throw an error every time any constrained variable is not considered as a covariate.

f-allian: @rsomers1998  lgtm then, feel free to merge.

f-allian (approved):

Issue #258, opened 2024-02-02
Spurious node when reading in causal DAG DOT files


jmafoster1:
**Describe the bug**
A spurious node called "\n" is created when we have a DAG specified as the following.
```
digraph dag {
  a -> b;
}
```
To avoid this, I have to manually remove the newline between the last edge and the closing brace, e.g.
```
digraph dag {
  a -> b;}
```
This is really annoying, especially when we have automatically generated DAGs. I think the problem is with pydot.

**To Reproduce**
Copy the above DAG into a dot file, then
```
causal_dag=CausalDAG("dag.dot")
print(causal_dag.graph.nodes())
```

**Expected behavior**
Nodes should be ["a", "b"]

**Observed behavior**
Nodes are ["a", "b", "\n"]

PR #268, opened 2024-03-12
Fix: Remove spurious nodes in dot files due to \n

f-allian:
Fixes issue #258 by:

1. Reading in the dot file,
2. Removing any new lines,
3. Passing the contents back into pydot.

### DIFF
--- a/causal_testing/specification/causal_dag.py
+++ b/causal_testing/specification/causal_dag.py
@@ -133,7 +133,9 @@ class CausalDAG(nx.DiGraph):
     def __init__(self, dot_path: str = None, **attr):
         super().__init__(**attr)
         if dot_path:
-            pydot_graph = pydot.graph_from_dot_file(dot_path)
+            with open(dot_path, 'r') as file:
+                dot_content = file.read().replace('\n', '')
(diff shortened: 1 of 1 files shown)


jmafoster1 (approved):
This is such a simple solution! I love it! I'm seeing a couple of pylint/black issues that might need looking at before merging, although I can't see what they are so they might be errors or something...

causal_testing/specification/causal_dag.py
```diff
@@ -133,7 +133,9 @@ class CausalDAG(nx.DiGraph):
     def __init__(self, dot_path: str = None, **attr):
         super().__init__(**attr)
         if dot_path:
-            pydot_graph = pydot.graph_from_dot_file(dot_path)
+            with open(dot_path, 'r', encoding='utf-8') as file:
+                dot_content = file.read().replace('\n', '')
```
cwild-UoS, line 137: I think it may be worth adding a comment describing this, as the problem this is solving is not obvious from the code.

cwild-UoS (approved):
Looks like a good solution to me

PR #269, opened 2024-03-18
Temporary workaround for "I(...) not in df" bug

jmafoster1:
Interaction terms within an `I(...)` block currently throw an error because the dmatrix sees these as atoms, so cannot pull variables from within them. I have implemented a temporary workaround for this which just ignores them, but we will need to get something better in the long term so that we can have categoricals within these blocks

### DIFF
--- a/causal_testing/testing/estimators.py
+++ b/causal_testing/testing/estimators.py
@@ -352,7 +352,14 @@ class LinearRegressionEstimator(Estimator):
         model = self._run_linear_regression()
         newline = "\n"
         patsy_md = ModelDesc.from_formula(self.treatment)
-        if any((self.df.dtypes[factor.name()] == 'object' for factor in patsy_md.rhs_termlist[1].factors)):
+        if any(
+            (
+                self.df.dtypes[factor.name()] == "object"
+                for factor in patsy_md.rhs_termlist[1].factors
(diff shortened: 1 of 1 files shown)


causal_testing/testing/estimators.py
```diff
@@ -352,7 +352,14 @@ def estimate_coefficient(self) -> tuple[pd.Series, list[pd.Series, pd.Series]]:
         model = self._run_linear_regression()
         newline = "\n"
         patsy_md = ModelDesc.from_formula(self.treatment)
-        if any((self.df.dtypes[factor.name()] == 'object' for factor in patsy_md.rhs_termlist[1].factors)):
+        if any(
```
f-allian, line 355: @jmafoster1 For my understanding, does this avoid the errors because you're checking if the terms already exist in the original dataframe? I'm happy to pass this for now, but can you send me a simple interaction term example when you have time so I can look into this?

f-allian (approved):

causal_testing/testing/estimators.py
jmafoster1: Yes, that's right. Thanks for approving. I will be publishing my replication package in a minute, which has the example in, so I'll send you a link

PR #282, opened 2024-07-10
Update python versions stated in docs

christopher-wild:


f-allian (approved):
@christopher-wild Thanks for updating these. Would it be better to just specify `Python >= 3.9`? Bit cleaner textually and will save us when future versions are released :)

PR #283, opened 2024-07-10
Remove conda from ci-tests

christopher-wild:


christopher-wild:
Chosen to remove Conda rather than fix it. I think our GH Action should use as many inbuilt features as possible and not introduce dependencies.

.github/workflows/ci-tests.yaml
```diff
@@ -20,25 +20,23 @@ jobs:
...
-      - name: Install package and dependencies
-        run: |
+    - uses: actions/checkout@v4
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: ${{ matrix.python }}
+    - name: Install dependencies
+      run: |
           python --version
```
f-allian, line 30: @christopher-wild  This isn't doing what it should be doing. Have a look at the [output logs](https://github.com/CITCOM-project/CausalTestingFramework/actions/runs/9887684997/job/27309948327) for e.g. Python 3.10 on Mac OS. Under install dependencies / test with PyTest, it's running the wrong version of python (3.12 in this case even though 3.10 is requested from `actions/setup-python@v5`.

Their [docs](https://github.com/actions/setup-python?tab=readme-ov-file#basic-usage) say:

> The python-version input is optional. If not supplied, the action will try to resolve the version from the default .python-version file. If the .python-version file doesn't exist Python or PyPy version from the PATH will be used. The default version of Python or PyPy in PATH varies between runners and can be changed unexpectedly so we recommend always setting Python version explicitly using the python-version or python-version-file inputs.

So it might be a simple fix by just changing `python-version: ${{ matrix.python }}` to `python-version: ${{ matrix.python-version }}`

f-allian (changes requested):
See comments below

f-allian (approved):
Nice!

PR #286, opened 2024-07-25
Temporal shenanigans

jmafoster1:
Added in temporal causal estimation with treatment sequences.


causal_testing/specification/capabilities.py
```diff
@@ -0,0 +1,75 @@
...
+            self.timesteps_per_intervention,
+            [(c.variable, c.value) for c in self.capabilities],
+        )
+        return strategy
+
+    def total_time(self):
+        """
+        Calculate the total duration of the treatment strategy.
+        """
+        return (len(self.capabilities) + 1) * self.timesteps_per_intervention
```
f-allian, line 75: @jmafoster1 Out of curiosity, why is the total time defined as `(n+1)*dt` instead of `n*dt`? Do you assume the treatment strategy starts at `t=dt` instead of `t=0`?

tests/specification_tests/test_capabilities.py
```diff
@@ -0,0 +1,42 @@
...
+    """
+    Test the Capability class for basic methods.
+    """
+
+    def setUp(self) -> None:
+        pass
+
+    def test_repr(self):
+        cap = Capability("v", 1, 0, 1)
+        self.assertEqual(str(cap), "(v, 1, 0-1)")
```
f-allian, line 20: @jmafoster1 Should this be `"(v, 1, 0, 1)"` instead of `"(v, 1, 0-1)"`?

tests/testing_tests/test_estimators.py
```diff
@@ -77,7 +79,7 @@ class TestLogisticRegressionEstimator(unittest.TestCase):
 
     @classmethod
     def setUpClass(cls) -> None:
-        cls.scarf_df = pd.read_csv("tests/data/scarf_data.csv")
+        cls.scarf_df = pd.read_csv("tests/resources/data/scarf_data.csv")
```
f-allian, line 82: @jmafoster1 Why did this change to `test/resources/data`? Do we have other subfolders in `tests/resources`?

f-allian (commented):
@jmafoster1 Looks good so far. Seems like a really seamless integration. I'm currently not on my laptop so I'll do further tests later this evening / tomorrow.

Also, it might be worth requesting a review from @christopher-wild if he has some time.

causal_testing/specification/capabilities.py
jmafoster1: @f-allian, yes exactly. This is slightly geared towards the water case study, where there is a grace period of dt before the first intervention takes place, and another period of dt after the last intervention before we can observe the final outcome. A slightly more general solution might be to have an extra `grace_period` parameter and then have the total time be `len(capabilities) * timesteps_per_intervention + grace_period`, but I'll keep it as it is for now.

tests/specification_tests/test_capabilities.py
jmafoster1: I thought the `0-1` notation for stringing capabilities would be easier to read, then you know that the last element represents the duration of the capability. Otherwise, I'd be constantly forgetting what the last three elements represented!

tests/testing_tests/test_estimators.py
jmafoster1: There was a folder called `tests/data` and one called `tests/resources/data`, both containing test data, so I thought I'd condense everything into one folder.

causal_testing/specification/capabilities.py
```diff
@@ -0,0 +1,75 @@
...
+"""
+from causal_testing.specification.variable import Variable
+
+
+class Capability:
+    """
+    Data class to encapsulate temporal interventions.
+    """
+
+    def __init__(self, variable: Variable, value: any, start_time: int, end_time: float):
```
christopher-wild, line 13: We should use the [Any](https://docs.python.org/3/library/typing.html#the-any-type) from the Typing library rather than the generic `any`. The reason for this is in theory the generic `any` isnt related to the concept of 'any' object and would fail if we had a static type checker

christopher-wild, line 13: Should start_time be `int` and end_time `float`?

causal_testing/testing/causal_test_adequacy.py
```diff
@@ -87,11 +94,21 @@ def measure_adequacy(self):
...
+            if self.group_by is not None:
+                ids = pd.Series(estimator.df[self.group_by].unique())
+                ids = ids.sample(len(ids), replace=True, random_state=i)
+                estimator.df = estimator.df[estimator.df[self.group_by].isin(ids)]
+            else:
+                estimator.df = estimator.df.sample(len(estimator.df), replace=True, random_state=i)
+            try:
+                results.append(self.test_case.execute_test(estimator, None))
+            except LinAlgError:
+                continue
```
christopher-wild, line 107: Would logging any warnings here be helpful? Is it okay if it fails to append results?

causal_testing/testing/estimators.py
```diff
@@ -599,3 +600,238 @@ def estimate_coefficient(self, bootstrap_size=100) -> tuple[pd.Series, list[pd.S
...
+            observations = individual.loc[
+                (individual["time"] % self.timesteps_per_intervention == 0) & (individual["time"] < fault_time)
+            ]
+            if not observations.empty:
+                fault_t_do.loc[observations.index[0]] = 1
+        assert sum(fault_t_do) <= 1, f"Multiple fault times for\n{individual}"
+
+        return pd.DataFrame({"fault_t_do": fault_t_do})
+
+    def setup_fault_time(self, individual, perturbation=-0.001):
```
christopher-wild, line 700: The typing for `individual` is missing. I only noticed as I wanted to see what kind of object you can invert with ~ and use `len` on

tests/specification_tests/test_capabilities.py
```diff
@@ -0,0 +1,42 @@
+import unittest
+from enum import Enum
```
christopher-wild, line 2: Is this import used?

causal_testing/testing/estimators.py
jmafoster1: Fixed. It's a pandas dataframe. You invert a boolean Series object with ~.

causal_testing/testing/causal_test_adequacy.py
jmafoster1: I've added in the warnings, but it's very OK to continue. What's happening here is that we're resampling the data and recalculating the estimate. If you don't have a lot of data, some of those samples may fail to converge, or fail in other ways to produce an estimate. This indicates that you don't have enough data, so I've now added "number of successes" as a third dimension to causal test adequacy.

causal_testing/specification/capabilities.py
jmafoster1: Good spot. Fixed.

jmafoster1: Fixed

jmafoster1:
If everyone's happy with this, please could one of you approve it so I can get it merged in?

christopher-wild:
> If everyone's happy with this, please could one of you approve it so I can get it merged in?

Ah sorry missed this! You can click re-request review next to ours name in the top right to send a notification if we do take too long

christopher-wild (approved):

jmafoster1:
Thanks @christopher-wild. Doing that feels very passive aggressive, though! :upside_down_face:

christopher-wild:
Haha I get what you mean but it's what it's there for 😁!

Issue #262, opened 2024-02-13
Check all variables in Patsy interaction term in estimate_coefficient method


cwild-UoS:
**Describe the bug**
If multiple variables are present in an interaction term (i.e. self.treatment = 'variable1:variable2'), each variable needs to be checked to be a type object. This is not currently implemented as it is not known how to return a list of variables from a Patsy formula neatly. 

**Additional context**
First look into if Patsy can return a list of variables somehow, if not manual string splitting and iterating over them.

jmafoster1:
I tried to run this with a more complex interaction term `driving_score ~ I(completion_score * (1 - outside_route_lanes))` and it doesn't like it because the `I` basically just puts everything inside it into a block, so patsy doesn't recognise to pull out `completion_score` and `outside_route_lanes` as terms. It just has `I(completion_score * (1 - outside_route_lanes))`.

cwild-UoS:
Not closed with #264,  PR changed scope

PR #290, opened 2024-08-07
GP formulae

jmafoster1:
The primary purpose of this PR is to add Genetic Programming to `LinearRegressionEstimator` class to infer more complex regression equations from the data, given the features identified from the DAG. For example, identification might give us an equation `Y ~ aX + bZ`. We could then use GP and the data to infer the relationship `Y ~ aX^2 + b log(Z)`. The features are not different, but the estimation will be much more accurate (and the equational form could be validated by the user).

In addition to this main functionality, I have taken the opportunity to refactor the estimation into a separate package, instead of all being in a single `estimators.py` file within `testing`. This is why there are several "new" files. I also had to fix a bunch of other files with the old `estimators.py` as a dependency.

TLDR; the main contributions are
- adding GP functionality to infer LR equations from data, given a list of features from causal identification.
- refactoring `estimators.py` to have each estimator be in its own separate file within an `estimation' package
- fixing test/demo code to reflect this


f-allian (commented):
@jmafoster1 Thanks for this Michael. Could you please add a few more bullet points on what changes this PR specifically includes? I can see there has been a lot added/refactored but I can't quite make sense of why / the benefits of them from your description above.

jmafoster1:
@f-allian sorry for the delay on this. I've been off ill. I've updated the initial comment at the top describing the changes.

causal_testing/estimation/regression_estimator.py
```diff
@@ -0,0 +1,119 @@
...
+        Define as `regressor = ...`` outside of __init__, not as `self.regressor = ...`, otherwise
+        you'll get an "cannot instantiate with abstract method" error.
+        """
+
+    def add_modelling_assumptions(self):
+        """
+        Add modelling assumptions to the estimator. This is a list of strings which list the modelling assumptions that
+        must hold if the resulting causal inference is to be considered valid.
+        """
+        self.modelling_assumptions.append(
```
jmafoster1, line 73: @f-allian , I am confused by this, since we instantiate several `LinearRegressionEstimator`s during testing, which is a subclass of this this that calls `super().__init__` (i.e. the method defined above), which itself calls its `super().__init__` method (from `estimator.py`), which calls `self.add_modelling_assumptions()`, so this line should be covered?

causal_testing/estimation/regression_estimator.py
```diff
@@ -0,0 +1,119 @@
...
+        x[self.treatment] = [self.treatment_value, self.control_value]
+
+        for k, v in adjustment_config.items():
+            x[k] = v
+        for k, v in self.effect_modifiers.items():
+            x[k] = v
+        x = dmatrix(self.formula.split("~")[1], x, return_type="dataframe")
+        for col in x:
+            if str(x.dtypes[col]) == "object":
+                x = pd.get_dummies(x, columns=[col], drop_first=True)
```
jmafoster1, line 115: It would be good to get this properly tested at some point, but it wasn't covered before. It's only flagging it now because I've moved the code around a sufficient amount that it thinks it's new (it's not). However, there's broader issues with this code that need addressing (see #262)

f-allian (approved):
@jmafoster1 I've fixed the code coverage warning (see below for details). 

Everything else looks fine overall, but it would be worth giving the estimation scripts more meaning filenames. For instance by changing `gp.py` to `genetic_programming_estimator.py` or `estimator.py` to `abstract_estimator.py` and so on. This would help with having a standardised naming convention and file structure.

Edit: I'll leave the merging / releasing the new major version with you :-)

PR #293, opened 2024-10-21
Amend Metamorphic Relations

f-allian:
- Added a simple logic to the function `generate_metamorphic_relations` in `specification/metamorphic_relation.py` to deal with situations with dags that contain cycles/loops.
- The added feature is a boolean parameter called `skip_ancestors`. The default is `False` which retains the original logic of the algorithm. If `True`, no adjustment sets are calculated and empty sets are passed into the `ShouldCase`/`ShouldNotCause` methods. 
- Logging is also used in the script's entry-point to warn the user to proceed with caution if this parameter is true.

TODO:

- [ ] Write unit tests to validate the functionality of the algorithm when `skip_ancestors=True`

### DIFF
--- a/causal_testing/specification/metamorphic_relation.py
+++ b/causal_testing/specification/metamorphic_relation.py
@@ -214,44 +214,63 @@ class MetamorphicTest:
         )
 
 
-def generate_metamorphic_relations(dag: CausalDAG) -> list[MetamorphicRelation]:
-    """Construct a list of metamorphic relations implied by the Causal DAG.
+def generate_metamorphic_relations(dag: CausalDAG, skip_ancestors: bool) -> list[MetamorphicRelation]:
+    """Construct a list of metamorphic relations based on the DAG or cyclic graph.
 
-    This list of metamorphic relations contains a ShouldCause relation for every edge, and a ShouldNotCause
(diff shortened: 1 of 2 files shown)


causal_testing/specification/metamorphic_relation.py
```diff
@@ -214,44 +214,63 @@ def __str__(self):
         )
 
 
-def generate_metamorphic_relations(dag: CausalDAG) -> list[MetamorphicRelation]:
-    """Construct a list of metamorphic relations implied by the Causal DAG.
+def generate_metamorphic_relations(dag: CausalDAG, skip_ancestors: bool) -> list[MetamorphicRelation]:
+    """Construct a list of metamorphic relations based on the DAG or cyclic graph.
```
rsomers1998, line 218: Unsure if actually relevant but in the literature a causal graph with loops is called a DCG (or you could just refer to either as a "causal graph". Pedantic suggestion though

jmafoster1:
I'm afraid I'm a little confused as to what's going on here. I was envisioning something along the lines of an `ignore_cycles` parameter which, if true, got all of the nodes that form part of any cycle and ignored (i.e. did not append it to the list of MRs) every test which had them as control, treatment, or a member of the adjustment set. I think simply ignoring the ancestors might not be strong enough.

PR #299, opened 2024-12-11
DAFNI Workflow Enhancements

f-allian:
Enhancements to the DAFNI workflow that enables better usability and reproducibility. These include:

- [x] Removed `variables.json` as an argument. The variables can now be stored as metadata attributes in the `dot` file produced by the user.
- [x] Developing a local test suite to ensure the entrypoint works as expected.
- [x] Turned `main_dafni.py` into a module, which improves the modularity and testability of the entrypoint in the test suite.
- [x] CI/CD workflows have also been updated.


dafni/README.md
```diff
@@ -10,8 +10,8 @@ to upload the framework onto [DAFNI](https://www.dafni.ac.uk).
 - `data` contains two sub-folders (the structure is important for DAFNI).
   - `inputs` is a folder that contains the input files that are (separately) uploaded to DAFNI.
     - `causal_tests.json` is a JSON file that contains the causal tests.
-    - `variables.json` is a JSON file that contains the variables and constraints to be used.
-    - `dag.dot` is a dot file that contains the directed acyclc graph (dag) file.
+    - `dag.dot` is a dot file that contains the directed acyclc graph (dag) file,
```
jmafoster1, line 13: Would it make sense to document the metadata attributes here?

dafni/README.md
```diff
@@ -10,8 +10,8 @@ to upload the framework onto [DAFNI](https://www.dafni.ac.uk).
 - `data` contains two sub-folders (the structure is important for DAFNI).
   - `inputs` is a folder that contains the input files that are (separately) uploaded to DAFNI.
     - `causal_tests.json` is a JSON file that contains the causal tests.
-    - `variables.json` is a JSON file that contains the variables and constraints to be used.
-    - `dag.dot` is a dot file that contains the directed acyclc graph (dag) file.
+    - `dag.dot` is a dot file that contains the directed acyclc graph (dag) file, 
+       including the variables (and constraints if necessary) as node attributes.
```
jmafoster1, line 14: How are constraints expressed as node attributes? I thought constraints came in the causal tests JSON file?

dafni/data/outputs/causal_tests_results.json
```diff
@@ -42,9 +48,15 @@
...
-      "ci_high": 213468.71609854023
+      "effect_estimate": {
+        "max_doses": 213111.93333333335
+      },
+      "ci_low": [
+        212755.15056812647
+      ],
+      "ci_high": [
+        213468.71609854023
+      ]
```
f-allian, line 59: @jmafoster1 The estimated values are consistent and haven't changed, which is good, but for some reason, it's storing them as a dictionary now. Do you know why it would suddenly do this?

dafni/README.md
f-allian: @jmafoster1 I was thinking of documenting this on the DAFNI platform itself, rather than in this README. Both?

dafni/src/main_dafni.py
```diff
@@ -91,53 +84,33 @@ def get_args(test_args=None) -> argparse.Namespace:
...
+        for node, attributes in causal_dag.graph.nodes(data=True)
+        if eval(attributes["typestring"]) == "input"
+    ]
 
-    with variables_path.open("r") as file:
-        inputs = json.load(file)
+    constraints = set()
 
-    return inputs
+    for (node, attributes), input_var in zip(causal_dag.graph.nodes(data=True), inputs):
```
jmafoster1, line 103: Expressing constraints in terms of individual variables is a real headache. I battled with this for about 6 months during my PhD. The problem is that constraints can relate variables, e.g. `i1 > i2`. Then you need to put the constraint `i1 > i2` for node `i1` and the inverse (i.e. `i2 < i1`) for node `i2`. (I guess you don't technically need to do this here, but then you have the question of which node you associate the constraint with.)

dafni/data/outputs/causal_tests_results.json
jmafoster1: Yes, Chris made this a thing ages ago in order to standardise the CTF outputs between numerical and categorical data. Categorical data will give you an effect estimate for each category of the variable. Doing it this way means we don't need to test datatypes everywhere.

dafni/src/main_dafni.py
jmafoster1: When I did the [CARLA case study](https://github.com/CITCOM-project/carla-case-study/blob/main/studied-cases/case-1/causal_tests.json) I encoded queries on a per-test basis using an attribute called "query", which is a silly name for it here but made sense at the time because you can then just do `df = df.query(test["query"])` and it'll "just work" if the user supplies the query in the correct format. I propose we do a similar thing here.

dafni/README.md
f-allian: @jmafoster1 I didn't realise `constraints` are defined in the causal tests file. In that case, we'll need something separate to parse the JSON file for the constraints. 

Do we have an example JSON file that I can use that uses constraints in the causal tests file? I can only find one in the Poisson example that [defines constraints in the script](https://github.com/CITCOM-project/CausalTestingFramework/blob/bb2527b01c1382e960b96ad637fce6e3efe3770e/examples/poisson/example_run_causal_tests.py#L118) but doesn't actually use it in the [JSON file](https://github.com/CITCOM-project/CausalTestingFramework/blob/main/examples/poisson/causal_tests.json) for some reason, which doesn't make sense to me.

dafni/src/main_dafni.py
```diff
@@ -91,53 +84,33 @@ def get_args(test_args=None) -> argparse.Namespace:
...
+    :param causal_dag: an instancce of the CausalDAG class containing the relationships and variables (as attribtutes)
     :returns:
-            - dict - A valid dictionary consisting of the causal tests
+    - Tuple containing the inputs, outputs and constraints to pass into the modelling scenario
     """
-    if not variables_path.exists() or variables_path.is_dir():
-        print(f"JSON file not found at the specified location: {variables_path}")
 
-        raise FileNotFoundError
+    inputs = [
```
jmafoster1, line 95: Since we just return the inputs and outputs as separate lists and then join them up later, would it make sense to just return one list of `variables` defined as
```
variables = [
        Input(node, eval(eval(attributes["datatype"]))) if attributes["typestring"] == input else
        Output(node, eval(eval(attributes["datatype"]))) if attributes["typestring"] == input
        for node, attributes in causal_dag.graph.nodes(data=True)
]
```
It's not really a problem either way, but it's just a bit more elegant

dafni/README.md
jmafoster1: I think both is good. I've been using the `main_dafni.py` as a main entrypoint for other things to save me having to write a separate run script each time, so it makes sense to document it for users who might not have access to the DAFNI platform.

dafni/README.md
jmafoster1: Yes, see https://github.com/CITCOM-project/CausalTestingFramework/pull/299#discussion_r1880022545

The constraints in the script are an unused variable, so I guess they weren't transfered to the JSON file. We probably put them in at some point because they were mandatory and then made them optional but didn't bother to get rid of the variable. This stuff was written a LONG time ago before we had linting for PRs, so I guess it got missed.

dafni/src/main_dafni.py
f-allian: @jmafoster1 I think it's less clear here what's being calculated and returned as a single line. You'd then have to define `inputs = variables[0]` which is less informative than the current method.

dafni/README.md
jmafoster1: In that case, would it make sense to have the dafni stuff be a separate repo? We may have had this discussion before, but if we only want the dafni stuff used on that platform, it probably doesn't make sense to distribute it with the CTF. Alternatively, a longer term goal could be to broaden the applicability of the entrypoint, but that doesn't invalidate this PR, so we can do that separately later.

dafni/src/main_dafni.py
jmafoster1: My point was that we never use `inputs` or `outputs` in isolation. We build them separately and return them separately, but it looks like the only thing we do with them is append them anyway, so I was wondering whether it made sense to treat them separately at all, but if you think it's more elegant the way you have it then we can leave it as is. It's not like people will be dealing with millions of inputs/outputs.

dafni/model_definition.yaml
```diff
@@ -20,6 +20,10 @@ metadata:
   contact_point_email: farhad.allian@sheffield.ac.uk
   source_code: https://github.com/CITCOM-project/CausalTestingFramework
   licence: https://github.com/CITCOM-project/CausalTestingFramework?tab=MIT-1-ov-file#readme
+  rights: If you use our model, please use the following instructions to cite our work - The paper citation should be the Causal Testing Framework paper, and the software citation should contain the specific Figshare DOI of the version used in your work.
+  subject: [ 'Biota', 'Boundaries', 'Climatology / Meteorology / Atmosphere', 'Economy', 'Elevation', 'Environment', 'Farming', 'Geoscientific Information', 'Health', 'Imagery / Base Maps / Earth Cover', 'Inland Waters', 'Intelligence / Military', 'Location', 'Oceans', 'Planning / Cadastre', 'Society', 'Structure', 'Transportation', 'Utilities / Communication' ]
```
jmafoster1, line 24: Where have all these tags come from? I'm not sure they're super relevant to the CTF...

f-allian: @jmafoster1 They're all of the available model tags on DAFNI. Since it's a general testing framework, it makes sense to tag it to all models available for users; thoughts?

jmafoster1: Not really sure what to suggest here. On the one hand, it should be appropriate to test all (or at least most) types of model. On the other hand, it does not model any of these things itself so, depending on what people are looking for, it might end up being "noise" in a lot of searches.

I also don't love giving it tags like "Military/Intelligence" as that might mean that infosec people then start to have more opinions about it! What happens if we don't give it any tags? Is that allowed?

f-allian: Good point - I'm unsure of tag(s) it's specifically relevant to, but it can be empty. Maybe I can request to make one for us? Alternatively, I can remove it.

jmafoster1 (approved):

PR #308, opened 2025-02-17
Jmafoster1/remove data collector

jmafoster1:
- `DataCollector` classes have been removed
- `ObservationalDataCollector` has been replaced by directly passing dataframes
- `ExperimentalDataCollector` has been replaced by `ExperimentalEstimator` class to directly run models
- Removed test generation and execution functionality from `MetamorphicRelation` since this can be done via the JSON front end
- Moved `MetamorphicRelation` classes into the `testing` package instead of `specification`
- Changed estimators (except `IPCWEstimator`) to take a `BaseTestCase` instead of a treatment and outcome variable


jmafoster1:
This PR looks to drop the overall percentage code coverage, but that's because I've ripped out a bunch of code, so the proportions have changed slightly.

jmafoster1:
Resolved merging conflicts from main

f-allian (changes requested):
@jmafoster1 Thanks Michael, this all looks good. However, we should also update our documentation to reflect these changes too. If you don't have the time let me know and I'll take it on 👍🏼

jmafoster1:
Good point. Done

f-allian:
@jmafoster1 There were still a few mentions of the data collector, and other errors/typos in the documentation so I've fixed those. But I've spotted a few other problems related to this PR.

- It appears that running `metamorphic_relations` generates causal test results wit 0 tests generated as an example on the dafni dataset, without any errors (see screenshot below). The expected output is of course a `json` file containing the causal tests. It looks like there are inconsistencies between how you've changed certain lines of code and how causal_base_test_case and causal_dag expect their input parameters, but unfortunately I don't know enough about their properties to fix this myself. We could probably rewrite this using a try/except block to catch this type of error better, and/or write tests that deals with such a case.

image

- Also, did you not say you wanted to remove `z3` and the `generation` folder? Are you leaving for another time? It'd make sense to do this in this PR too.

jmafoster1:
> It appears that running metamorphic_relations generates causal test results wit 0 tests generated as an example on the dafni dataset

This is because there's now an explicit `test` attribute on each node in the DAG which needs to be set to `True` for a test to be generated. It looks like I've been generating them as False (or missing them off entirely in a couple of cases for that case study). I might change the default behaviour to only ignore tests where it's explicitly set to `False` for backwards compatibility.

jmafoster1:
> Also, did you not say you wanted to remove z3 and the generation folder?

One thing at a time! I was going to save that for a separate PR, but I guess we can do that here. As for the generation folder, it _can_ go if necessary, but I'm not as desperate to get rid of it as the data collector classes. It probably will go though since it never really amounted to anything.

f-allian:
> This is because there's now an explicit `test` attribute on each node in the DAG which needs to be set to `True` for a test to be generated. 

Where? I can't find any test attributes in `metamorphic_relation.py`. Also, there are a lot of type hints that don't correspond to the inputs/outputs so it'd be worth fixing that.

> but I guess we can do that here. As for the generation folder, it can go if necessary, but I'm not as desperate to get rid of it as the data collector classes. It probably will go though since it never really amounted to anything.

Yeah it makes sense to do that in this PR. I'll remove these.

jmafoster1:
> Where? I can't find any test attributes in metamorphic_relation.py.

Line 208 where we generate `nodes_to_test`. I changed the default behaviour so that only nodes which have `test=False` will not be tested (rather than only testing nodes with `test=True`).

jmafoster1:
> Also, there are a lot of type hints that don't correspond to the inputs/outputs so it'd be worth fixing that.

Please could you handle that one?

> Yeah it makes sense to do that in this PR. I'll remove these.

I've just been looking into this. It's actually turning out to be a bit complicated because it's quite baked into the JSON frontend. I'm thinking it might be worth taking this opportunity to rebuild the json frontend as the main entry point rather than just a utility class that needs to be instantiated.

jmafoster1:
Once I'd got rid of the causal test suite and the abstract causal test case classes, stripping out the rest of the Z3 references was actually pretty straightforward. One thing we do need to be careful of now though is that the `scenario.constraints` currently doesn't interact with the data at all (i.e. we're not doing `df.query` anywhere) because `Scenario` objects are separate from causal test cases. This will need to be done in the entrypoint. To be honest, I quite often skip creating them altogether, but let's not strip them out too!

f-allian:
@jmafoster1 This looks fine now. There are still some mentions of the data collector / JSON class in our `examples/` directory, but it makes sense to leave that for another PR once we've had our discussions. 👍🏼

f-allian (approved):

jmafoster1:
No need to re-review. Once this batch of tests pass, I'll get this PR merged in and do the JOSS release

Issue #303, opened 2025-01-15
Create a Main Entry Point to the CTF


f-allian:
- After removing the legacy JSON front-end and the causal-test-suite, we require a main entry point to call the CTF

f-allian:
Closed by #309 and #316.

Issue #315, opened 2025-03-06
Functionality needed for test-specific queries


f-allian:
- It'd be useful for users to have the option to include test-specific queries to filter the data in addition to the global filtering.

f-allian:
Closed by #316

PR #316, opened 2025-03-06
Functionality for test-specific queries

f-allian:
- The entry point can now also parse test-specific queries to filter the data. For example, users can configure their `causal_tests.json` file to include a query for each test in the following format, e.g.:

```
{
  "tests": [
    {
      "name": "Temperature Test",
      "estimator": "LinearRegressionEstimator",
      "estimate_type": "coefficient",
      "effect": "direct",
      "treatment_variable": "humidity",
      "query": "temperature > 25",
      "expected_effect": {
        "plant_growth": "Positive"
      },
      "formula": "plant_growth ~ humidity + sunlight + soil_quality",
      "alpha": 0.05,
      "skip": false
    },
  ]
}
```

- This means users now have the ability to filter globally on the runtime data using the optional `--query` flag, or they can, again optionally, configure test-specific queries for each test.

- Closes issue #303 and sub-issue #315


causal_testing/main.py
```diff
@@ -281,17 +281,26 @@ def create_causal_test(self, test: dict, base_test: BaseTestCase) -> CausalTestC
         if estimator_class is None:
             raise ValueError(f"Unknown estimator: {test['estimator']}")
 
+        # Filter the data if a test-specific query exists, otherwise use the original data
+        test_query = test.get("query", self.query)
```
jmafoster1, line 285: Please could this be conjoined rather than overriding? I can imagine a use case where someone has a general query (i.e. a modelling scenario) and wants a test specific query too (i.e. accessing a particular value in the CATE). Such a causal question might be "How does closing schools affect the number of old people with COVID for modelling scenarios in the UK?", where "modelling scenarios in the UK" is the general query applied to every test case and "old people" is the test specific query. You might then also have causal tests for children, young adults, and middle aged folks. I actually did this in my CARLA case study, except I didn't need a modelling scenario query here since I could control my dataset to only contain relevant data.

jmafoster1 (changes requested):

jmafoster1 (approved):
Thanks, looks good

PR #322, opened 2025-05-06
Update example_vaccine.py

Jimazcoder:
modified causal test case and linear regression estimator to fit class attributes


jmafoster1 (approved):
Thanks @Jimazcoder, our tests should have picked this up.

PR #324, opened 2025-05-16
Batch test execution

jmafoster1:
For the EA case study, trying to execute 87000 test cases fills up the RAM. The `run_tests_in_batches` method was meant to fix this, but didn't. I really want to be able to run all 87K tests at once, so I dug a bit deeper into this. The problem is that, when we execute a test case, the `estimator.model` attribute becomes non-None. It is this that fills up the RAM, but we don't use it here, so I just reset it to None after the test is executed. I'm not sure how I feel about this as a solution going forward, so any better suggestions would be much appreciated. I think `test.estimator.model` is used elsewhere, though (perhaps as part of test adequacy?) but we should be able to do everything we need to do before resetting it to None. On the other hand, it just feels a bit inelegant somehow.


f-allian:
@jmafoster1 I agree, definitely not the most elegant solution.

If it is because of the `test_case.estimator.model` attribute, it looks like it's coming from the [abstract `RegressionEstimator` class](https://github.com/CITCOM-project/CausalTestingFramework/blob/15513be06b295acd165a74492f6e66f92752bd62/causal_testing/estimation/abstract_regression_estimator.py#L89) where `self.model` is defined and later used for the prediction stages. 

Can you explain a few things for me first, you say:

> The problem is that, when we execute a test case, the estimator.model attribute becomes non-None. It is this that fills up the RAM, but we don't use it here, so I just reset it to None after the test is executed

1. Isn't it that the test case can be executed because the `estimator.model` attribute isn't non-None? 
2. Why isn't the `estimator.model` used here? It looks like it's needed to run the regression algorithm on the input data.
3. Using this hacky method of setting `estimator.model` to `None`, how much time does it actually save you compared to not doing it?
4. This isn't too important but I'm curious - what batch sizes have you tested?

jmafoster1:
1 and 2. Yes and no. Of course, we need the model to evaluate the test case, but we don't need it to be assigned to `test_case.estimator` to do that. The basic process at the moment is
1. Train the model
2. Use this to evaluate the test
3. Assign it to the instance of `estimator` so it can be used later if needed (we don't need this here)

3. The problem here is that I physically can't run all 87000 tests in one hit without this. What I'd been doing before was breaking the causal tests into two batches, running them separately, and concatenating the results. This has a comparable runtime, but is annoying as I have to run 3 things instead of just one and have to babysit it a bit more (and discover through trial and error how many batches I need). With this, I can just press go and come back when it's done, so it's a lot easier.

4. I set the default to 100, which is completely arbitrary. That seems to use negligible extra RAM and runs acceptably fast, so I left it at that.

jmafoster1:
@f-allian I've removed the `estimator.model` property. It was only being used in tests, so I've either removed redundant assertions or modified them accordingly. There was only one that really needed the statsmodels model explicitly, which was for Richard's validation stuff, so I've made the `train_model` method public so that it can just be accessed directly.

f-allian (approved):
@jmafoster1 Excellent, nice one Michael!

jmafoster1:
Yeah I think so. I must have missed your review or something

Issue #327, opened 2025-06-02
NetworkX 3.5 compatibilty: d_separated function deprecation


f-allian:
## Problem
- CI tests (drafts) are failing with `AttributeError: module 'networkx' has no attribute 'd_separated'` after the NetworkX 3.5 release.
- The function's been changed from `d_separated` to `is_d_separated`  (see [previous versions](https://networkx.org/documentation/networkx-3.2.1/reference/algorithms/generated/networkx.algorithms.d_separation.d_separated.html#networkx.algorithms.d_separation.d_separated) vs. [latest version](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.d_separation.is_d_separator.html#networkx.algorithms.d_separation.is_d_separator))
- Note: this was likely overlooked in `ci-tests-drafts.yml` compared to our main `ci-tests.yml` file because of caching of versions in its workflow file, so I'll fix that too.

## Fix 

- Replace all `nx.d_separated` calls with `nx.is_d_separator`.

f-allian:
@jmafoster1 For your visibility on this - NetworkX have also dropped Python 3.9/3.10 support in v3.5 (see their [documentation](https://networkx.org/documentation/stable/install.html)), but I really don't think we should drop support for our codebase because of one line of code change in their API. So I'm going to just enforce 3.4 <= NetworkX < 3.5 for us. This'll also avoid any breaking changes/major releases for now. 

Does this sound okay to you?

f-allian:
Closing this for now but it might be good to return to this in the future and discuss our forward compatibility.

PR #328, opened 2025-06-02
Enforce NetworkX 3.4

f-allian:
Quick fix to #327

### DIFF
--- a/tests/testing_tests/test_causal_test_case.py
+++ b/tests/testing_tests/test_causal_test_case.py
@@ -206,31 +206,42 @@ class TestCausalTestExecution(unittest.TestCase):
 
     def test_estimate_params_none(self):
         """Check that estimate_params defaults to empty dict when None is passed into the estimator object"""
+        estimator = LinearRegressionEstimator(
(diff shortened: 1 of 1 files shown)


causal_testing/testing/causal_test_case.py
```diff
@@ -47,6 +47,10 @@ def __init__(
         self.estimator = estimator
         if estimate_params is None:
             self.estimate_params = {}
+
+        else:
+            self.estimate_params = estimate_params
```
jmafoster1, line 52: @f-allian please could you add a couple of tests for this? Maybe one explicitly setting it to empty and one setting a parameter for something (e.g. "formula" for the `LinearRegressionEstimator` class?

f-allian: @jmafoster1 Good spot - sure.

jmafoster1 (approved):
I modified the assertions a bit on the tests so that the test case is actually run. I realised that the `formula` parameter wasn't actually a valid estimate parameter for `estimate_ate`, so wanted to test the parameters were actually getting used properly.

PR #330, opened 2025-06-13
Moved test generation from DAG files into main functionality

jmafoster1:
Now users can type `python -m causal_testing -G...` to generate tests rather than `python causal_testing/testing/metamorphic_relation.py`.


f-allian (approved):
@jmafoster1 All looks good! 👍🏼

PR #336, opened 2025-07-09
Validation to check that outcome and treatment aren't the same

### LINKED ISSUE #114: JSON Frontend, add validation to check that outcome and treatment aren't the same

jmafoster1:
Closes #114. Added validation in `BaseTestCase` to check that treatment and outcome are not the same.

### DIFF
--- a/causal_testing/testing/base_test_case.py
+++ b/causal_testing/testing/base_test_case.py
@@ -14,6 +14,9 @@ class BaseTestCase:
     :param effect: A string representing the effect, current support effects are 'direct' and 'total'
     """
 
-    treatment_variable: Variable
-    outcome_variable: Variable
-    effect: str = Effect.TOTAL.value
+    def __init__(self, treatment_variable: Variable, outcome_variable: Variable, effect: str = Effect.TOTAL.value):
+        if treatment_variable == outcome_variable:
+            raise ValueError(f"Treatment variable {treatment_variable} cannot also be the outcome.")
+        self.treatment_variable = treatment_variable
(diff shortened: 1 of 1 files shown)


causal_testing/testing/base_test_case.py
```diff
@@ -17,3 +17,7 @@ class BaseTestCase:
     treatment_variable: Variable
     outcome_variable: Variable
     effect: str = Effect.TOTAL.value
+
+    def __post_init__(self):
+        if self.treatment_variable == self.outcome_variable:
+            raise ValueError(f"Treatment variable {self.treatment_variable} cannot also be the outcome variable.")
```
f-allian, line 23: @jmafoster1 For my benefit, what's this `__post_init__ method`? I've never come across it before.

f-allian (approved):

causal_testing/testing/base_test_case.py
jmafoster1: It's a [special dataclass method](https://docs.python.org/3/library/dataclasses.html#dataclasses.__post_init__) that is called at the end of `__init__`. Because it's a frozen dataclass, we can't define an `__init__` method like we normally would (it gives lots of errors saying we're not allowed to), so this allows us to perform the validation as part of `__init__` under the hood. It was a suggestion from Gemini, so feel free to suggest an alternative solution.

f-allian: @jmafoster1 Not at all - that's great. I've just never seen the naming convention before. I'll update with main and merge now then

PR #337, opened 2025-07-11
Experimental optimisations, proposed by ChatGPT.

neilwalkinshaw:
### Prompt: "Can you speed this up?", pasting CausalDag.enumerate_minimal_adjustment_sets

Summary of changes:

moral_graph[t] instead of nx.neighbors(...)
_Accesses adjacency dict directly, faster in memory_

update(...) instead of union with comprehension
_Avoids building intermediate sets_

Eliminated intermediate list conversions
_Avoids unnecessary copies_

Variable renaming (pbd_graph, ancestor_graph, etc.)
_Improves clarity and prevents redundant calls_

### Same applied to list_all_min_sep

Summary of changes:

graph[node] instead of nx.neighbors(graph, node)
_Faster adjacency access in memory_

treatment_component = ... break
_Avoids unnecessary iteration after finding the component_

sample(sorted(...), 1)[0]
_Returns a value, not a set, avoiding later unpacking_

Removed repeated set(...) wrapping
_Reduces GC and memory allocations_

### Same applied to constructive_backdoor_criterion

Avoids repeated set(self.nodes)
_self.nodes is probably already a set-like iterable_

Combines descendant updates efficiently
_Reduces overhead of set.union with unpacking_

Avoids constructing logger message unless needed
_Significant savings if logging level > INFO_

Clearer and faster condition check with & (set intersection)
_Cleaner than issubset(difference(...))_


jmafoster1:
I've been through this. Most of the optimisations were fine. Some were a little odd. I also took the opportunity to add some extra ones of my own. Also, most notably, I made this class a _proper_ subclass of `nx.DiGraph` so we don't have to keep accessing `.graph` all the time. It doesn't make it any faster, but it does make it much cleaner.

PR #346, opened 2025-07-30
Changed main run command to facilitate test generation

### LINKED ISSUE #338: Help not showing with -h option
If you run `python -m causal-testing -h`, it gives an error message saying that you've not specified the DAG, data, etc. We should be able to just show the help message without having to do this.

### LINKED ISSUE #341: Make default estimator configurable
From test generation

jmafoster1:
I changed the top level entrypoint of the CTF to add `generate` and `test` commands so that test generation from DAGs can also be done at the top level. Closes #338 by allowing the `--help` option to display properly. Additionally, closes #341 by making the default estimator for the `generate` command configurable.


f-allian (approved):
@jmafoster1 LGTM 👍🏼 I'll have to do another big doc revamp as part of #326

Issue #343, opened 2025-07-14
You need visual studio to install the CTF on Windows


jmafoster1:
I don't think you should have to do that, but it seems like you need a C++ compiler to install numpy.

jmafoster1:
This is apparently insurmountable as it's a prerequisite of Scipy. We will, however, provide a docker image, which should hopefully make this a less awful process for windows users.

f-allian:
Closing this issue for now as I can't replicate the error:

```
Using pip 25.2 from C:\Users\cs1fxa\PycharmProjects\test\.venv-312\Lib\site-packages\pip (python 3.12)
Collecting causal-testing-framework
  Obtaining dependency information for causal-testing-framework from https://files.pythonhosted.org/packages/b3/2e/e95e523f102e989cdffc9101086fdb52047865c703af126e1f4bdad65e9d/causal_testing_framework-11.0.0-py3-none-any.whl.metadata
  Downloading causal_testing_framework-11.0.0-py3-none-any.whl.metadata (13 kB)
Collecting fitter~=1.7 (from causal-testing-framework)
  Obtaining dependency information for fitter~=1.7 from https://files.pythonhosted.org/packages/f9/20/e0526cff78b376e74f9e153d66b946650068bc9ff24864023a85ec963f99/fitter-1.7.1-py3-none-any.whl.metadata
  Downloading fitter-1.7.1-py3-none-any.whl.metadata (8.2 kB)
Collecting lifelines~=0.29.0 (from causal-testing-framework)
  Obtaining dependency information for lifelines~=0.29.0 from https://files.pythonhosted.org/packages/fa/7d/b227ca1bae2474f954f00644dcf692f75701043717565a1a56685cb69416/lifelines-0.29.0-py3-none-any.whl.metadata
  Downloading lifelines-0.29.0-py3-none-any.whl.metadata (3.2 kB)
Collecting lhsmdu~=1.1 (from causal-testing-framework)
  Obtaining dependency information for lhsmdu~=1.1 from https://files.pythonhosted.org/packages/a2/5c/c0f4e69093a4ebfc52c34cb1926dbe5f29181fcb934a624c4123d6ce1921/lhsmdu-1.1-py2.py3-none-any.whl.metadata
  Downloading lhsmdu-1.1-py2.py3-none-any.whl.metadata (5.7 kB)
Collecting networkx<3.5,>=3.4 (from causal-testing-framework)
[...]

PR #348, opened 2025-07-30
Precompile Wheels for Different OS and Python Versions Before Publishing to PyPI

### LINKED ISSUE #347: Build platform-specific wheels to avoid dependency compilation issues on Windows
Although our framework is written in pure Python, it depends on several libraries with compiled components (e.g., `numpy`, `scipy`, `pyarrow`). Users installing the package, especially on Windows, may encounter build failures or be required to install `Microsoft Visual C++ Build Tools`, if pre-built wheels for these dependencies are unavailable or pip falls back to source builds.

To avoid these issues and improve installation reliability across platforms, we should:

- Integrate [cibuildwheel](https://github.com/pypa/cibuildwheel) into the GitHub Actions workflow.

- Build and publish platform-specific wheels for Linux, Windows, and macOS.

- Ensure all binary dependencies are pre-bundled during the build process to eliminate the need for compilation at install time.

This will avoid any unnecessary C/C++ compilers by default.

f-allian:
- Fixes to our CD pipeline. The workflow now builds and publishes Python wheels for multiple operating systems and Python versions.
- It uses `cibuildwheel` to generate wheels for Ubuntu, macOS (including Apple Silicon arm64), and Windows, supporting Python 3.10, 3.11, and 3.12. 
- Source distribution is included for completeness. 
- All artifacts are collected and published to PyPI, ensuring all modern CPU architectures are covered too.
- Ultimately, users don't need to have compilers (such as Microsoft C++) installed to compile the CTF. Closes #347 and #343

### DIFF
--- a/.github/workflows/publish-to-pypi.yaml
+++ b/.github/workflows/publish-to-pypi.yaml
@@ -1,4 +1,4 @@
-name: Publish python PyPI
+name: Publish to PyPI
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


jmafoster1 (approved):
Thanks Farhad, hopefully this will resolve the Windows installation issues.

PR #353, opened 2025-08-07
EffectEstimate dataclass

### LINKED ISSUE #340: Confidence intervals are lists
We should make them dictionaries to be consistent with the point estimates.

jmafoster1:
Closes #340.
- Renamed `TestValue` class to `EffectEstimate` and moved it into `estimation`
- Changed to having estimators return an EffectEstimate instance rather than a tuple

This is much cleaner than the old solution as estimation methods now return one tidy object that's always of the same type.


f-allian (approved):
@jmafoster1 Good idea moving this all into a data class. Looks good to me 👍🏼

PR #380, opened 2026-03-31
Update readthedocs ubuntu version

jmafoster1:
ReadTheDocs announced that the deprecation of the Ubuntu 20.04 LTS build image. This PR updates the workflow to use 24.04.


f-allian (approved):
Keeping note of this - we might have to return to this PR in the future as I remember Python 3.11 being problematic for RTD but we won't know this until it's been deployed.

PR #382, opened 2026-04-29
Jmafoster1/329 estimator endpoint

### LINKED ISSUE #329: Enable Custom Estimators in Entrypoint
- Our main entry-point currently only supports `LinearRegression` and `LogisticRegression` estimators by directly hardcoding them in.
- We should enable a factory/registry to support custom estimator classes that can be programmatically registered and reference by name in test configuration files. 
- This is important as we have already have the base abstract estimator classes, and some examples in the `examples` directory rely on custom classes.

jmafoster1:
Closes #329. This PR makes estimators and expected causal effects endpoints instead of having to manually convert via a dict.

### DIFF
--- a/causal_testing/main.py
+++ b/causal_testing/main.py
@@ -273,27 +273,17 @@ class CausalTestingFramework:
                 "pyproject.toml file."
             )
 
-        # Handle combined queries (global and test-specific)
-        test_query = test.get("query")
-        combined_query = None
-
-        if self.query and test_query:
-            combined_query = f"({self.query}) and ({test_query})"
-            logger.info(
-                f"Combining global query '{self.query}' with test-specific query "
-                f"'{test_query}' for test '{test['name']}'"
-            )
-        elif test_query:
-            combined_query = test_query
-            logger.info(f"Using test-specific query for '{test['name']}': {test_query}")
-        elif self.query:
-            combined_query = self.query
-            logger.info(f"Using global query for '{test['name']}': {self.query}")
-
-        filtered_df = self.data.query(combined_query) if combined_query else self.data
+        # Handle global queries
+        # Test-specific queries are handled by the estimator as not all estimators support them
+        filtered_df = self.data
+        if self.query:
+            filtered_df = self.data.query(self.query)
 
         # Create the estimator with correct parameters
         estimator_class = estimator_map.get(test["estimator"]).load()
+        estimator_kwargs = test.get("estimator_kwargs", {})
+        if "query" in test:
+            estimator_kwargs["query"] = test["query"]
         estimator = estimator_class(
             base_test_case=base_test,
             treatment_value=test.get("treatment_value"),
@@ -303,11 +293,8 @@ class CausalTestingFramework:
                 self.dag.identification(base_test, self.scenario.hidden_variables()),
             ),
             df=filtered_df,
-            effect_modifiers=None,
-            formula=test.get("formula"),
             alpha=test.get("alpha", 0.05),
-            query=combined_query,
-            **test.get("estimator_kwargs", {}),
+            **estimator_kwargs,
         )
 
         # Get effect type and create expected effect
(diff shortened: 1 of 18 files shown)


causal_testing/main.py
```diff
@@ -259,48 +257,30 @@ def create_causal_test(self, test: dict, base_test: BaseTestCase) -> CausalTestC
...
+                f"Unsupported estimator {test['estimator']}. Supported: {sorted(estimator_map)}. "
+                "If you have implemented a custom estimator, you will need to add this to your entrypoints via your "
+                "pyproject.toml file."
             )
-        elif test_query:
-            combined_query = test_query
-            logger.info(f"Using test-specific query for '{test['name']}': {test_query}")
-        elif self.query:
-            combined_query = self.query
-            logger.info(f"Using global query for '{test['name']}': {self.query}")
```
f-allian, line 300: @jmafoster1 Are we dropping functionality for test-specific queries now?

tests/resources/data/tests.json
```diff
@@ -4,19 +4,23 @@
...
     "estimate_type": "coefficient",
-    "effect_modifiers": [],
-    "query": "test_input > 0",
     "expected_effect": {"test_output": "NoEffect"},
-    "skip": false
+    "skip": false,
+    "estimator_kwargs": {
+      "query": "test_input > 0",
+      "effect_modifiers": []
+    }
```
f-allian, line 12: @jmafoster1 Are these parameters required now? What happens if the user doesn't supply `estimator_kwargs`?

pyproject.toml
```diff
@@ -65,6 +65,23 @@ Documentation = "https://causal-testing-framework.readthedocs.io/"
...
+InstrumentalVariableEstimator = "causal_testing.estimation.instrumental_variable_estimator:InstrumentalVariableEstimator"
+IPCWEstimator = "causal_testing.estimation.ipcw_estimator:IPCWEstimator"
+
+[project.entry-points."causal_effects"]
+NoEffect = "causal_testing.testing.causal_effect:NoEffect"
+SomeEffect = "causal_testing.testing.causal_effect:SomeEffect"
+Positive = "causal_testing.testing.causal_effect:Positive"
+Negative = "causal_testing.testing.causal_effect:Negative"
+ExactValue = "causal_testing.testing.causal_effect:ExactValue"
+
```
f-allian, line 84: @jmafoster1 I've read the Python entrypoint documentation you shared earlier, but I haven't seen this approach being used before. 

1. Does this register the in-built estimator classes we have?
2. What would the general workflow look like for users? Do they have to `pip install -e . ` every time they create a new custom estimator / causal_effect?

f-allian (approved):
@jmafoster1 Generally good to go. I'll do some stress testing later but I just had a few questions for my own benefit!

causal_testing/main.py
jmafoster1: No, the estimators all handle it anyway, so there's no need to do it here as well. The alternative would be to move the "query" argument from the estimators, which I could also do. That's possibly a bit DRY-er, but I'm not sure whether it might be nice to keep it as an option if people want to use the estimators without driving it from the main frontend?

tests/resources/data/tests.json
jmafoster1: Not required, but they should be supplied in that format _if they are being supplied_. Nothing would happen if they weren't supplied, just no extra arguments other than the already required ones would be passed

jmafoster1: I chose this way because some of the estimators have different keyword arguments that people might want to supply (e.g. `instrument` for the IV estimator) and some don't take the `formula` keyword, so we don't want to be passing it in as we did before.

pyproject.toml
jmafoster1: 1. Yes
2. Yes, but they don't need to add them to the CTF pyproject.toml file. They can add the entrypoints to their own pyproject.toml file, e.g. if they were testing their own project, they could implement the custom estimator there, add the entrypoint to their pyproject.toml file, and the standard `pip install causal_testing_framework` would still work for them. I've used this on another project and it all seems to magically "just work". You do need to `pip install -e .` the project of whatever pyproject.toml file you modify, but that's not a huge problem, and it does open up a massive flexibility of use

f-allian: @jmafoster1 I think for no. 2 if it'd be beneficial for users for us to mention this in the docs, e.g. under the `estimators.rst` file you made: "Note: you don't need to add your entrypoints to your pyproject.toml, but if you do ... ensure you reinstall the framework using `pip install -e .`"

jmafoster1: Good shout. Added

PR #385, opened 2026-06-19
SURE - Discovering causal structure

Jake248Newman:


causal_testing/discovery/hill_climber.py
```diff
@@ -0,0 +1,236 @@
...
+    individual.graph["time"] = round(end_time - start_time)
+    if output_file is not None:
+        nx.drawing.nx_agraph.write_dot(
+            individual,
+            output_file,
+        )
+    return individual
+
+
+if __name__ == "__main__":
```
jmafoster1: Please can you drop this `main` part out? The main interaction should come from the endpoint (and should be the user supplying their own data).

causal_testing/__main__.py
```diff
@@ -39,6 +41,17 @@ def main() -> None:
     # Setup logging
     setup_logging(args.verbose)
 
+    if args.command == Command.DISCOVER:
+        logging.info("Discovering causal structures")
```
jmafoster1:
```suggestion
        logging.info("Discovering causal structure")
```

causal_testing/main.py
```diff
@@ -567,6 +593,14 @@ def parse_args(args: Optional[Sequence[str]] = None) -> argparse.Namespace:
         help="Run tests in batches of the specified size (default: 0, which means no batching)",
     )
 
+    #Discovery
+    parser_discover = subparsers.add_parser(Command.DISCOVER.value, help="Discover causal structures from data")
+    parser_discover.add_argument("-d", "--data-path", help="Path to data file (.csv)", required=True)
```
jmafoster1:
```suggestion
    parser_discover.add_argument("-d", "--data-paths", help="Paths to data files (.csv)", nargs="+", required=True)
```
This makes the argument consistent with the testing parser, and also allows us to support multiple data files (i.e. to stack together data with the same column headings scattered across several files. Of course, you'll need to use `pd.concat` to concatenate them before you use them

jmafoster1 (commented):
This is looking good so far!

tests/discovery_tests/test_hill_climber.py
```diff
@@ -0,0 +1,97 @@
...
+        mock_simple_cycle.side_effect = [cycle_edges, []]
+        
+        remove_cycles(mock_dag, included_edges=set())
+        assert mock_dag.remove_edge.call_count == 1
+        call_args = mock_dag.remove_edge.call_args[0]
+        assert call_args in cycle_edges
+        mock_dag.add_nodes_from.assert_called_once_with(['A', 'B', 'C'])
+
+    @patch('causal_testing.discovery.hill_climber.simple_cycle')
+    def test_remove_cycles_respects_included_edges(self, mock_simple_cycle):
```
jmafoster1: Do we need to use mocking here? It might be more intuitive to just set up a DAG, unless there's some specific functionality we need the mock to bypass?

Jake248Newman: I think it was useful for how the function worked before, but no, it works well without it now

docs/source/modules/discovery.rst
```diff
@@ -0,0 +1,41 @@
...
+
+.. note::
+   Automated causal discovery is a starting point. The resulting DAG must always be manually inspected to ensure it 
+   is a valid representation of your system.
+
+Configuration Options
+---------------------
+The tool supports various configurations to tailor the discovery process to your specific dataset and domain knowledge:
+
+* **Fitness Functions** Choose between score-based and multi-objective fitness functions.
```
jmafoster1: I've dropped out the ability to specify different fitness functions on the Hill Climber, since what I was actually thinking of was the multi-objective NSGA that I've now added.

docs/source/modules/discovery.rst
```diff
@@ -0,0 +1,41 @@
...
+
+* **Search Constraints (Iterations)**
```
jmafoster1: Probably worth documenting each technique and its parameters separately here. We can potentially use `:autoclass` for this, like I did [here](https://github.com/test-flare/pytest-flakefighters/blob/main/docs/source/configuration.rst)

docs/source/modules/discovery.rst
```diff
@@ -0,0 +1,41 @@
...
+  .. container:: zoom-container
+
+     .. image:: ../../../examples/covasim_/vaccinating_elderly/dag.png
+        :class: zoomable-image
+        :alt: Causal DAG of the vaccinating the elderly modelling scenario
+
+  * **Excluding Edges:** If domain knowledge dictates that ``max_doses`` has no causal effect on any outputs, you 
+    can specify ``max_doses -> ".*"`` and ``".*" -> max_doses`` in the *exclude edges* dot file.
+  * **Including Edges:** If it is known that ``vaccine`` directly affects all three outputs, you can specify 
+    ``vaccine -> "cum_.*"`` in the *include edges* dot file.
```
jmafoster1: It'd also be good to include the exact command to run to get the DAG (and possibly a quick tutorial IPYNB) in `tutorials/`

jmafoster1 (approved):
Thanks for your work on this @Jake248Newman. This looks fine to me, so feel free to merge today, unless you can see anything obviously missing. We can make separate PRs for any minor changes we need as part of the evaluation

PR #395, opened 2026-07-16
Jmafoster1/388 fix large scenarios

### LINKED ISSUE #388: Large DAGs/data take a prohibitive amount of memory
**Describe the bug**
Generating tests for a large DAG/dataset takes a lot of memory. I think this is because a new estimator is defined for each test case, where each estimator has its own copy of the dataset. I think it would be better to have tests as a generator.

**To Reproduce**
1. Take a large causal DAG (50+ nodes)
2. Take a large dataset (10k+ rows)
3. Try to generate causal test cases
4. Watch your memory fill up until the process is killed, or you computer becomes non-responsive.

**Expected behavior**
The CTF should handle larger scenarios smoothly.

jmafoster1:
Closes #388.
This has turned into a bit of a monster PR. Main changes as follows:
- Estimators no longer store the dataframe. This dramatically reduces the amount of memory taken up causal test cases.
- The `CausalTestingFramework` class is now a "first class citizen". It is designed and intended to be used as a standalone API rather than purely as part of the commandline interface.
- `CausalTestCase`s now have some extra instance attributes:
  - They now store the result of `execute_test` as an instance variable instead of returning it. This means you don't have to iterate a `zip` object of the tests and their associated results.
  - They now store the `name` and `skip` attributes from the JSON frontend. This means that there's now a 1:1 mapping between tests and their representation in JSON, meaning that we don't need to keep the parsed input around for later use as part of output.


causal_testing/__main__.py
```diff
@@ -60,9 +205,14 @@ def main() -> None:
...
             # Otherwise you end up with duplicate indices, which causes problems further down the line
-            df = pd.concat([pd.read_csv(path) for path in args.data_paths]).reset_index()
+            df = pd.concat([read_dataframe(path) for path in args.data_paths]).reset_index()
             if args.variables:
                 df = df[args.variables]
+            # Drop unnamed columns
+            unnamed_columns = [c for c in df.columns if c.startswith("Unnamed: ")]
+            if unnamed_columns:
+                logger.warning(f"Dropping unnamed columns: {unnamed_columns}")
+            df = df.loc[:, ~df.columns.str.contains("^Unnamed: ")]
```
jwhogg: I'd use `df = df.drop(columns=unnamed_columns)`  inside the if, as you are currently searching twice

causal_testing/causal_testing_framework.py
```diff
@@ -0,0 +1,335 @@
...
+        """
+        Create causal test case from test configuration and base test.
+
+        :param test: Dictionary containing test configuration parameters
+
+        :return: CausalTestCase object
+        :raises: ValueError if invalid estimator or configuration is provided
+        """
+        estimator_map = {ff.name: ff for ff in entry_points(group="estimators")}
+        effect_map = {ff.name: ff for ff in entry_points(group="causal_effects")}
```
jwhogg, line 197: How big is this list comp? is it efficient to be using a dict here?

jmafoster1: They're both pretty small, like 5 elements. The entrypoints are defined in the corresponding entries in pyproject.toml. What would you propose to use instead?

jwhogg: if they are small can leave it as is

causal_testing/estimation/linear_regression_estimator.py
```diff
@@ -92,27 +62,28 @@ def gp_formula(
         formula = gp.run_gp(ngen=ngen, pop_size=pop_size, num_offspring=num_offspring, seeds=seeds)
         formula = gp.simplify(formula)
         self.formula = f"{self.base_test_case.outcome_variable.name} ~ I({formula}) - 1"
-        self.setup_covariates()
+        self._setup_covariates(df)
```
jwhogg, line 65: what does adding the underscore do here?

jmafoster1: Conventionally, it makes it a private function, so folks don't (or at least shouldn't) try to call it themselves outside the module. I think I've come up with a better solution for this that will alleviate the "test adequacy is prohibitively slow to calculate" issue, but I'll leave that for a separate PR

jwhogg (approved):
lgtm

Issue #399, opened 2026-07-20
ExactValue causal effect doesn't work for categorical treatments


jmafoster1:
Gives `ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()` when you try to evaluate it.

Issue #391, opened 2026-07-16
Do we need the `MetamorphicRelation` class anymore?


jmafoster1:
I think the functionality that these provide could be provided more simply via the `CausalTestCase` class, especially since the main thing we want to do with the metamorphic relations is turn them into causal test cases so we can execute them...

Issue #387, opened 2026-07-10
Improve test "generate and run" API


jmafoster1:
Current API expects a 2 step process in which users first generate tests from a DAG, save these to JSON, and then load in data, tests, and DAG in a separate process. Increasingly, I'm finding that I just load in DAG and data and want to generate and run tests from just that. We should try to improve this workflow. In particular, I think we should get rid of the `CausalTestingPaths` class and revamp the `setup` method in `main.py`.

PR #401, opened 2026-07-21
Jmafoster1/dag evaluation

### LINKED ISSUE #393: Generate causal test cases directly from DAG
At the moment, we have to generate a list of `MetamorphicRelation` objects, convert them to JSON, and then load that in as causal test cases. It'd be good to be able to go straight from DAG to causal test suite.

jmafoster1:
Closes #393 #399 #391 #387.

- Removes the `MetamorphicRelation` class in favour of generating causal tests directly.
- Improved `to_dict` methods for easier serialisation.
- Improved "generate and run tests" pipeline from DAG
- Added `EVALUATE` command to evaluate a causal DAG against a dataset.


jmafoster1:
Codecov isn't happy here because the codebase has got smaller. I have removed quite a lot of code, and thus the tests that previously covered that code. If you try to inspect the difference in coverage, it says "No Files covered by tests were changed". Hopefully this will go away once #395 is in.

jmafoster1:
Subsumed by #402

Issue #394, opened 2026-07-16
Do we need the `BaseTestCase` class?


jmafoster1:
This is just a data class that is never used outside of the `CausalTestCase` class. If I remember rightly, it was set up to reduce the number of instance variables of a `CausalTestCase` to satisfy pylint, but then we further increased the number of instance variables, so pylint is sad again anyway. Personally, I think we should just have `CausalTestCase` and get rid of the `BaseTestCase`.

Issue #392, opened 2026-07-16
Do we need the `Scenario` or `Variable` classes anymore?


jmafoster1:
The `Variable` classes (`Input`, `Output`, `Meta`) were made to support test data generation (i.e. to generate inputs for the system), but this never really came to fruition. Since then, they've just hung around without providing much functionality. We could probably just use string names now, since we never differentiate between the different variable types. If we ever did want to provide different functionality based on variable type (e.g. input and output), then this could be recorded as metadata in the causal DAG (both in the DOT file representation and in the `CausalDAG` class).

The `Scenario` class was made to hold the `Variable`s and the "modelling scenario": a query on the data, e.g. "location = UK". While I do think it's important to keep the query functionality at both the DAG level (so the user doesn't have to pre-filter their data) and the edge level (to allow for low-effort adjustment of categorical variables), I think we can manage this more easily through the new improved `CausalTestingFramework` class.

PR #402, opened 2026-07-21
Jmafoster1/marie kondo

### LINKED ISSUE #398: Do we need `effect_modifiers` and `adjustment_config`?
Both are dictionaries mapping variables to values. They're always handled in the same way together, so we can probably just merge them into one variable and then document "This is how you handle effect modifiers", which most people probably never will anyway...

jmafoster1:
Closes #398 #394 #392

This PR gets rid of a load of relic infrastructure classes that no longer [spark joy](https://konmari.com/marie-kondo-rules-of-tidying-sparks-joy/). The result is a much more streamlined testing workflow. Notable removals include:
- the `MetamorphicRelation` class, in favour of directly generating test cases from a causal DAG
- The `Scenario`, `Variable`, and `BaseTestCase` classes, in favour of expressing variables as strings, and modelling scenarios as pandas query strings.
- The `CubicSplineEstimator` class, in favour of specifying cubic splines as formulas with `LinearRegressionEstimator`.
- Treatment and outcome variables are now specified as part of the estimator only.

It also adds a new command "evaluate" to evaluate a DAG with respect to a given dataset (with confidence intervals for the number of passing test cases calculated by resampling, similar to data adequacy).


jmafoster1:
OK, so looking at the `CausalTestCase` class, it seems that the `treatment_variable` and `outcome_variable` properties aren't ever actually used (except in serialising to/from JSON), so we could get rid of them altogether and just store them in the associated `Estimator`. Programmatically this makes sense, but intuitively I'm not keen. The vibes are off - it puts too much responsibility on the "estimator". However, since the estimator also stores the adjustment set and (in the case of regression), the equational relationship, I think it might make sense to rename the `Estimator` family to something like `CausalRelationship` or something from the world of structural equational models. What are other people's thoughts?

For reference, instantiating a causal test case currently looks like this:
```
estimation_model = LinearRegressionEstimator(
            treatment_variable="A",
            outcome_variable="C",
            treatment_value=self.treatment_value,
            control_value=self.control_value,
            adjustment_set=self.minimal_adjustment_set,
        )
        causal_test_case = CausalTestCase(
            treatment_variable="A",
            outcome_variable="C",
            expected_causal_effect=self.expected_causal_effect,
            estimator=estimation_model,
            effect_measure="ate",
        )
```

In the new system, it'd look like this, which would be a lot cleaner:
```
causal_relationship = LinearRegressionRelationship(
            treatment_variable="A",
            outcome_variable="C",
            treatment_value=self.treatment_value,
            control_value=self.control_value,
            adjustment_set=self.minimal_adjustment_set,
        )
        causal_test_case = CausalTestCase(
            expected_causal_effect=self.expected_causal_effect,
            causal_relationship=causal_relationship,
            effect_measure="ate",
        )
```

jmafoster1:
Update: Turns out removing `treatment_variable` and `outcome_variable` from `CausalTestCase` was really easy as nothing depended on it! I've added in `treatment_variable` and `outcome_variable` property methods for backwards compatibility and easy access to the variables, but the specification of test cases is a lot cleaner now. I think I will rename the `Estimator` classes to `CausalRelationship` to create a better intuition of what they do. This will essentially allow us to create full structural causal models!

causal_testing/estimation/abstract_regression_estimator.py
```diff
@@ -23,40 +22,86 @@ class RegressionEstimator(Estimator):
...
             self.formula = formula
+            self._adjustment_set_from_formula()
+            if adjustment_set is not None and set(adjustment_set) != set(self.adjustment_set):
+                raise ValueError(
+                    f"Specified formula {self.formula} does not match specified adjustment set {adjustment_set}"
+                )
+        elif adjustment_set is not None:
+            self.adjustment_set = adjustment_set
+            terms = [treatment_variable] + sorted(list(adjustment_set))
+            self.formula = f"{outcome_variable} ~ {' + '.join(terms)}"
```
jwhogg, line 52: in the `_adjustment_set_from_formula`, doing sorted() turns it into a list I beleive, is there a reason you need it as a sorted list?

and doing `self.adjustment_set = adjustment_set` will set it to a list globally.

jmafoster1: Good spot, this is probably a layover from something else. I'll revert it to being a set

jwhogg (approved):
looks good, pending that `set` change

PR #403, opened 2026-07-31
Jmafoster1/visualisation

jmafoster1:
This PR adds the capability to visualise causal test results on the DAG.

## Main changes
- Moved visualisation code out of `abstract_discovery.py` and into a separate dedicated visualisation module, which can grow sustainably over time
- Moved `effect_direction` into `CausalTestResult` class. Ideally, this would be part of `EffectEstimate`, but that leads to circular imports with `CausalEffect`


causal_testing/testing/causal_effect.py
```diff
@@ -45,7 +45,7 @@ def apply(self, effect_estimate: EffectEstimate) -> bool:
         else:
             raise ValueError(f"Test Value type {effect_estimate.type} is not valid for this CausalEffect")
 
-        return (~((effect_estimate.ci_low <= value_to_check) & (value_to_check <= effect_estimate.ci_high))).all()
+        return (~((effect_estimate.ci_low <= value_to_check) & (value_to_check <= effect_estimate.ci_high))).any()
```
jmafoster1, line 48: Bug fix: This should have been `.any()` all along

PR #405, opened 2026-08-03
Little fixes to facilitate the sharr evaluation

jmafoster1:
# Main changes
`hill_climber_discovery.py`
- Normalising fitness by number of generated tests
- Using a `for` loop for generational process instead of a `while`

'causal_dag.py`
- Override `add_edges_from` to include a check for cycles
- Removed resulting dead code cycle check from constructor

`__main__.py`
Outputting raw discovered DAG rather than a pretty rendering of passing/failing test results.

Tests for the above.


jmafoster1:
Looking at the codecov report, there's not actually anything missing, it's just that the codebase has got slightly smaller.

jwhogg (approved):
lgtm

